PR #102, opened 2024-08-21
Performance improvements

mokimo:
There are a few discussions about [preloading](https://github.com/orgs/adobecom/discussions/2719), as well as after this initial research, I'll open a discussion based on the advantages of parallelizing loading.

As project unity will only display POST-LCP, we can parallelize as much as the browser allows us to do, as we will need the assets anyway. I tried to keep this PR "unintrusive" by not changing any order of JS operations and simply moving a few imports and loading actions around.

### Observable improvements:
I've used local overwrites to test this on the prod pages behind akamai:
- Prod before, desktop: 1.73s
- Prod after, desktop: 1.31s

- Prod before, fast 4G (mobile): 3.34s
- Prod after, fast 4G (mobile) 2.13s

There is a bit of jitter within this data, but there should be general improvements. I'd love to see us observing the improvements using [CRUM](https://adobe-mwp.slack.com/archives/C06FV8K0PH8/p1723851457534279) and collect valuable data about this use-case. Numbers tested locally are usually _optimistic_ (e.g. 1.73s or 1.31s both sound too good for desktop) - so I'd love to follow those numbers.

### Future enhancement suggestions
- Method 1: Load a "facade" of the project unity so you don't even need most the logic components, JS, SVGs, CSS until _after_ the LCP and the base buttons will do.
- Method 2: Load project unity **after** the LCP (this might need some styling changes to reserve the necessary space to avoid CLS)

The general aim here is, we want to show the marquee (or any) image as fast as we can. Everything that is not part of the LCP block // image can be delayed. The user will not _instantly_ click the buttons within the marquee and even if an event listener to detect the click and loading the rest of the logic is good enough (gnav does that)

### Deep dive into what changed

Let’s establish the baseline, we queue a lot of imports. Each import // load needs to `(a)wait` for the response to be returned:
1-combine-load-before

1 - We can combine loading styles & workflow:
Screenshot 2024-08-17 at 07 46 42

2 - Move loading product workflow together with the base workflow 

3-combine-load-workflows

3 - Require workflow step assets (upload-bin, app-connector) earlier within the product workflow, to start loading them whenever the product workflow has been loaded

4-normal-import-for-required-assets

4 - Loading all assets (svgs) concurrently and re-use them. IMO We can be even smarter here in the future by only loading the SVGs that are initially visible, we might end up loading too much. But this PR should serve as a good inspiration and good first batch of improvements.

5-parallelize-PS-WF

**Test URLs:**
- Before: https://stage--unity--adobecom.hlx.page/drafts/ruchika/remove-background?martech=off
- After: https://perf-improvements--unity--adobecom.hlx.page/drafts/ruchika/remove-background?martech=off

### DIFF
--- a/unitylibs/blocks/unity/unity.js
+++ b/unitylibs/blocks/unity/unity.js
@@ -16,14 +16,37 @@ function getUnityLibs(prodLibs, project = 'unity') {
   return libs;
 }
 
+function getWorkFlowInformation(el) {
+  let wfName = '';
+  const workflowCfg = {
+    'workflow-photoshop': {
+      removebg: { endpoint: 'providers/PhotoshopRemoveBackground' },
+      changebg: { endpoint: 'providers/PhotoshopChangeBackground' },
+      slider: {},
+    },
+  };
+  [...el.classList].forEach((cn) => { if (cn.match('workflow-')) wfName = cn; });
+  if (!wfName || !workflowCfg[wfName]) return [];
+  return [wfName, workflowCfg[wfName]];
+}
+
 export default async function init(el) {
   const projectName = 'unity';
   const unitylibs = getUnityLibs('/unitylibs', projectName);
-  const [{ default: wfinit }] = await Promise.all([
+  const [wfName, wfDetail] = getWorkFlowInformation(el);
+  const [{ default: wfinit }, { default: productWfInit }] = await Promise.all([
     import(`${unitylibs}/core/workflow/workflow.js`),
+    import(`${unitylibs}/core/workflow/${wfName}/${wfName}.js`),
     new Promise((resolve) => {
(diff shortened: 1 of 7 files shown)


mokimo:
Here's the 
Prod before: 1.73s
prod 1 73 before desktop

Prod after: 1.31s
prod 1 31 after desktop

Prod before mobile: 3.34s
prod fast 4g 3 34 - mobile

Prod after mobile: 2.13s
prod fast 4g 2 13mobile

aishwaryamathuria:
Closing this PR as we are migrating to V2 and the performance has been handle as a part of https://github.com/adobecom/unity/pull/205
@dstrong23 @Ruchika4 @sanjayms01 @mokimo

PR #103, opened 2024-08-30
MWPW-157171 Building skeleton for Acrobat Workflow

sanjayms01:


unitylibs/core/workflow/workflow-acrobat/target-config.json
```diff
@@ -0,0 +1,18 @@
+{
+  "acom-widget": {
+    "verb": "fillsign",
```
aishwaryamathuria, line 3: We should get the verb as actions from the block itself .... Like :removebg: or any other action in place of hardcoding in the config to make unity independent of the block

unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -0,0 +1,151 @@
...
+  //         cOpts.payload.operations[idx][op.adjustmentType] = parseInt(op.filterValue.sliderElem.value, 10);
+  //       }
+  //     }
+  //   });
+  //   this.serviceHandler.postCallToService(
+  //     this.psApiConfig.connectorApiEndPoint,
+  //     { body: JSON.stringify(cOpts) },
+  //   );
+  // }
+}
```
aishwaryamathuria, line 151: As you mentined in the comments on slack we would need cleanup and anyways this will need more development to integrate with backend service.
But can we remove the comments to make this more readable?

unitylibs/core/workflow/workflow-acrobat/target-config.json
sanjayms01: Got it, will update accordingly.

PR #105, opened 2024-09-03
MWPW-157171 Building skeleton for Acrobat Workflow

sanjayms01:


sanjayms01:
@aishwaryamathuria @Ruchika4 Please let me know if this looks fine so I can merge this in for us to continue further with development. Selectors in config will need to change but will leave them as is for now. Thanks.

PR #106, opened 2024-09-05
Acrobat fill and sign upload functionality

Ruchika4:
* Upload a pdf from UI
* Create chunks to upload on temp urls
* Connector api call for acrobat.

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -35,7 +35,7 @@ export default class ActionBinder {
     );
     this.serviceHandler = new ServiceHandler(
       this.workflowCfg.targetCfg.renderWidget,
-      this.canvasArea
+      this.canvasArea,
     );
     for (const value of values) {
       switch (true) {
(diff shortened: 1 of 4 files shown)


unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -66,6 +66,82 @@ export default class ActionBinder {
...
+          [assetId]: {
+            name: filename,
+            size: filesize,
+            type: filetype,
+          }
+        },
+      },
+    };
+    const response = await this.serviceHandler.postCallToService(
+      this.acrobatApiConfig.connectorApiEndPoint,
```
sanjayms01, line 137: Is this defined somewhere? Should this be added into a new `this.acrobatApiConfig.acrobatEndpoint.connector` key value pair?

unitylibs/core/workflow/workflow-acrobat/target-config.json
```diff
@@ -1,12 +1,17 @@
 {
   "acom-widget": {
```
sanjayms01, line 2: Think we can go ahead and update this to `acom-widget.fillsign`.

Ruchika4: Not updating this as of now I see Blaine's comment and we might need to do more changes in config as the block name is changed

unitylibs/core/workflow/workflow-acrobat/action-binder.js
Ruchika4: As discussed marking this as resolved

PR #131, opened 2024-10-14
MWPW-160156 -Fix page loading issue on safari

Ruchika4:
fix issue of page with unity not loading on some versions of safari like 15, 16 etc.

Resolves: [MWPW-160156](https://jira.corp.adobe.com/browse/MWPW-160156)

Test URLs:

Before: https://stage--dc--adobecom.hlx.page/acrobat/online/test/sign-pdf-projectunity?unitylibs=stage
After: https://stage--dc--adobecom.hlx.page/acrobat/online/test/sign-pdf-projectunity?unitylibs=MWPW160156

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -326,7 +326,7 @@ export default class ActionBinder {
       };
       const finalizeJson = await this.serviceHandler.postCallToService(
         this.acrobatApiConfig.acrobatEndpoint.finalizeAsset,
-        { body: JSON.stringify(finalAssetData), signal: AbortSignal.timeout(15000)},
+        { body: JSON.stringify(finalAssetData), signal: AbortSignal.timeout?.(15000)},
       );
       if (!finalizeJson || Object.keys(finalizeJson).length !== 0) {
         await this.showSplashScreen();
--- a/unitylibs/scripts/utils.js
+++ b/unitylibs/scripts/utils.js
@@ -128,7 +128,7 @@ export async function priorityLoad(parr) {
   const promiseArr = [];
   parr.forEach((p) => {
     if (p.endsWith('.js')) {
-      const pr = loadScript(p, 'module', 'async');
+      const pr = loadScript(p, 'module', { mode: 'async' });
       promiseArr.push(pr);
     } else if (p.endsWith('.css')) {
       const pr = new Promise((res) => { loadLink(p, { rel: 'stylesheet', callback: res }); });
--- a/unitylibs/scripts/utils.js
+++ b/unitylibs/scripts/utils.js
@@ -124,14 +124,54 @@ export async function createActionBtn(btnCfg, btnClass, iconAsImg = false, swapO
   return actionBtn;
 }
 
+export function loadLinks(href, { as, callback, crossorigin, rel, fetchpriority } = {}) {
+  return new Promise((resolve, reject) => {
+    let element;
+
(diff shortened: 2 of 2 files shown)


aishwaryamathuria (approved):

narcis-radu:
@Ruchika4 - I'm not convinced this is the way to go, we are diverging from Milo with this local `loadLinks` implementation. I see the `loadLink` code is working fine on Milo pages, why is there a problem with Unity? Would it be possible to adjust the code so that it uses `loadLink` or `loadScript` and `loadStyle` from Milo ?

Ruchika4:
> https://stage--dc--adobecom.hlx.page/acrobat/online/test/sign-pdf-projectunity?unitylibs=stage

Hi @narcis-radu , thanks for the review. Basically to fix the safari issue I am adding js file as a script tag in head instead of link. And I did thought of using loadScript function available in milo and that also resolves the issue because as mentioned above it adds the js file in a script tag in head. However it doesn't have an option to add async or defer in the script tag ([Milo Code](https://github.com/adobecom/milo/blob/ec48f3c105aef6f8b1682cbae35788ba099902ec/libs/utils/utils.js#L398)) hence loading of this script will block the page from rendering as per my understanding. And I was trying to avoid that for performance reasons. I can update milo loadScript function too to add async parameter if it is passed when calling loadScript function. Please let me know if that approach sounds better.

narcis-radu:
@overmyheadandbody , @mokimo , @vhargrave , @robert-bogos - what do you think about @Ruchika4's suggestion? Would this be something that could be used by Milo / consumers or should we just isolate to Unity? Thanks!

overmyheadandbody:
But Milo's `loadScript` could be called at the point you'd need it, right? So you wouldn't be blocking something from the beginning. The method resembles https://github.com/adobecom/milo/blob/ec48f3c105aef6f8b1682cbae35788ba099902ec/libs/utils/utils.js#L322 quite a lot. For consistency purposes, my vote would also be to make updates to that method, should they be needed in the first place. The power of Milo lies in easily sharing such utilities and keeping them in sync across consumers, breaking that pattern will likely lead to issues later on.

mokimo:
+1 on @narcis-radu & @overmyheadandbody point to use milo native methods and checking if the milo method needs to be updated, if it faces the same page loading issues.

Ruchika4:
> But Milo's `loadScript` could be called at the point you'd need it, right? So you wouldn't be blocking something from the beginning. The method resembles https://github.com/adobecom/milo/blob/ec48f3c105aef6f8b1682cbae35788ba099902ec/libs/utils/utils.js#L322 quite a lot. For consistency purposes, my vote would also be to make updates to that method or the one you references, should they be needed. The power of Milo lies in easily sharing such utilities and keeping them in sync across consumers, breaking that pattern will likely lead to issues later on.
> 
> As @narcis-radu mentioned, it is surprising that we haven't seen this issue in Milo. Is this a very recent issue that we should be checking on Milo pages as well? Could the culprit be somewhere else?

We are seeing the safari issue when we are are trying to convert the output of loadLink function in a promise and try to resolve it in Promise.all() for a js file. I am not aware if there is a similar use case in any other milo consuming site though

Ruchika4:
I have updated the PR to use milo functions and have also raised a small PR to milo to update loadScript function here https://github.com/adobecom/unity/pull/131
Please review. Thanks

cc @narcis-radu  @mokimo  @overmyheadandbody

mokimo (approved):

sanjayms01 (approved):

robert-bogos (approved):
Thank you for addressing the feedback @Ruchika4 👍

PR #150, opened 2024-11-13
[MWPW-159324] Progress loader circle delay handling

robert-bogos:
**Description:**
This PR handles the unwanted delay in the loader circle render by moving the async loading elements outside of the function which gets triggered by clicking the "Remove background" or "Upload image" button.

**Resolves:** [MWPW-159324](https://jira.corp.adobe.com/browse/MWPW-159324)

**Testing instructions:**
Do these for both the before and after test URLs:
1. open an incognito window.
2. go to the test url.
3. try clicking the "Remove background" or "Upload image" button.
4. Observe the delay between the click and the loader circle render.

**Test URLs:**
Before: https://stage--unity--adobecom.hlx.page/drafts/rbogos/remove-background
After: https://mwpw-159324-loader-delay--unity--adobecom.hlx.page/drafts/rbogos/remove-background

### DIFF
--- a/unitylibs/core/workflow/workflow-photoshop/action-binder.js
+++ b/unitylibs/core/workflow/workflow-photoshop/action-binder.js
@@ -24,6 +24,26 @@ export default class ActionBinder {
     this.errorToastEl = null;
     this.psApiConfig = this.getPsApiConfig();
     this.serviceHandler = null;
+    this.loadServiceHandler();
+    this.loadSVGsMaybe();
+  }
+
+  async loadServiceHandler() {
+    const { default: ServiceHandler } = await import(`${getUnityLibs()}/core/workflow/${this.workflowCfg.name}/service-handler.js`);
+    this.serviceHandler = new ServiceHandler(
+      this.workflowCfg.targetCfg.renderWidget,
+      this.canvasArea,
+    );
+  }
+  
+  async loadSVGsMaybe() {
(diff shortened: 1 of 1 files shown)


aishwaryamathuria (approved):

unitylibs/core/workflow/workflow-photoshop/action-binder.js
```diff
@@ -24,6 +24,26 @@ export default class ActionBinder {
     this.errorToastEl = null;
     this.psApiConfig = this.getPsApiConfig();
     this.serviceHandler = null;
+    this.loadServiceHandler();
+    this.loadSVGsMaybe();
```
aishwaryamathuria, line 28: This will add both progress circle and all the svgs (Including the ones that are present on the buttons that will be made visible only when user interacts) in the critical rendering path. Can we split this into two functions and call the progress circle js and css loader here and load svg in psActions?

robert-bogos: Maybe I'm missing something, but this will not work because we'll still load the other SVGs when the user clicks on "Remove background". 
This will still cause a delay between the button click and the `'show'` action for the progress circle.

aishwaryamathuria:
Closing this PR as the changes are integrated and will be pushed as a part of https://github.com/adobecom/unity/pull/205
@dstrong23 @robert-bogos

PR #153, opened 2024-11-18
MWPW-162658 Adding enhanced LANA logging

sanjayms01:


Ruchika4:
Hi @sanjayms01 , I see we are no longer passing e.showError to dispatchErrorToast function. We had added that so we don't throw error if user clicks cancel button in splash screen.
It's value was set here https://github.com/adobecom/unity/blob/MWPW-162658/unitylibs/core/workflow/workflow-acrobat/action-binder.js#L265

sanjayms01:
> Hi @sanjayms01 , I see we are no longer passing e.showError to dispatchErrorToast function. We had added that so we don't throw error if user clicks cancel button in splash screen. It's value was set here https://github.com/adobecom/unity/blob/MWPW-162658/unitylibs/core/workflow/workflow-acrobat/action-binder.js#L265

Thanks for calling that out. This has been added back.

joaquinrivero (approved):

Ruchika4 (approved):

PR #155, opened 2024-11-20
MWPW-162832- Pass locale code to prelude

Ruchika4:
- Pass locale code for locale mapping at prelude side

Resolves: [MWPW-162832](https://jira.corp.adobe.com/browse/MWPW-162832)

Test URLs:

Before: https://main--cc--adobecom.hlx.page/drafts/ruchika/remove-background1?unitylibs=stage
After: https://main--cc--adobecom.hlx.page/drafts/ruchika/remove-background1?unitylibs=locale-code

### DIFF
--- a/unitylibs/scripts/utils.js
+++ b/unitylibs/scripts/utils.js
@@ -68,7 +68,7 @@ export function getLocale() {
   const urlLocale = window.location.pathname.split('/')[1];
   const { locales } = getConfig();
   const locale = Object.entries(locales).find(([key]) => key === urlLocale);
-  return locale ? locale[1] : 'us';
+  return locale ? locale[0] : 'us';
 }
 
 export async function loadSvg(src) {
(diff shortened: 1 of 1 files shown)


sanjayms01 (approved):

unitylibs/scripts/utils.js
```diff
@@ -68,7 +68,7 @@ export function getLocale() {
   const urlLocale = window.location.pathname.split('/')[1];
   const { locales } = getConfig();
   const locale = Object.entries(locales).find(([key]) => key === urlLocale);
-  return locale ? locale[1].ietf : 'en-US';
+  return locale ? locale[0] : 'us';
```
aishwaryamathuria, line 71: Hi @Ruchika4 In place of splitting the url to find the locale and parsing through the object could we use the getConfig().locale? This property returned by config has the correct entry of the lcoale from the current url and can return the required property

aishwaryamathuria (changes requested):
Hi @Ruchika4 In place of splitting the url to find the locale and parsing through the object could we use the getConfig().locale? This property returned by config has the correct entry of the locale from the current url and can return the required property

Ruchika4:
> Hi @Ruchika4 In place of splitting the url to find the locale and parsing through the object could we use the getConfig().locale? This property returned by config has the correct entry of the locale from the current url and can return the required details.

Code has been updated. Please check.

aishwaryamathuria (approved):

PR #156, opened 2024-11-20
MWPW-155682 Add dynamic import for errors.js

sanjayms01:
[Performance Optimization] Adding dynamic import for getError method for dispatching errors.

Resolves: [MWPW-155682](https://jira.corp.adobe.com/browse/MWPW-155682)

Test URLs:

Before: https://stage--dc--adobecom.hlx.page/acrobat/online/test/sign-pdf?unitylibs=stage
After: https://stage--dc--adobecom.hlx.page/acrobat/online/test/sign-pdf?unitylibs=MWPW-155682-errors-import

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -11,7 +11,6 @@ import {
   loadArea,
   loadImg,
 } from '../../../scripts/utils.js';
-import getError from '../../../scripts/errors.js';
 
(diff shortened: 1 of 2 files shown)


unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -159,7 +158,10 @@ export default class ActionBinder {
     if (showError) {
```
aishwaryamathuria, line 158: nit: If (!showError) return;
# early return

PR #162, opened 2024-11-22
MWPW-162943-Fix button text not visible issue on tab

Ruchika4:
- Fix button text not visible issue on tab

Resolves: [MWPW-162943](https://jira.corp.adobe.com/browse/MWPW-162943)

Test URLs:

Before: https://stage--cc--adobecom.hlx.page/products/photoshop/ai?unitylibs=stage
After: https://stage--cc--adobecom.hlx.page/products/photoshop/ai?unitylibs=MWPW-162943

### DIFF
--- a/unitylibs/core/styles/styles.css
+++ b/unitylibs/core/styles/styles.css
@@ -505,6 +505,10 @@
     color: var(--color-white);
   }
 
+  .unity-enabled .interactive-area.light .unity-action-area .unity-action-btn:focus-visible {
+    color: var(--color-black);
+  }
+
   .continue-in-app-button {
     background-color: var(--highlight-blue);
   }


sanjayms01 (approved):

robert-bogos (commented):
The code looks good, but, using the before URL, the CTA text does not disappear on focus 🤔
It gets the `color: var(--color-black);` from [here](https://github.com/adobecom/unity/blob/stage/unitylibs/core/styles/styles.css#L49).

My question is if this is still an issue? so that we don't push a fix for a problem that does not exist.

Ruchika4:
> The code looks good, but, using the before URL, the CTA text does not disappear on focus 🤔 It gets the `color: var(--color-black);` from [here](https://github.com/adobecom/unity/blob/stage/unitylibs/core/styles/styles.css#L49).
> 
> My question is if this is still an issue? so that we don't push a fix for a problem that does not exist.

Hi @robert-bogos , I am still able to reproduce the error
Screen Shot 2024-11-25 at 3 31 57 PM

sigadamvenkata:
issue tested with give branch and now tabbing focused on button CTA and can see the CTA test as well. PFA
{85EC6FCC-E0F2-47DE-BAD7-9ECFD67352F8}

robert-bogos:
@Ruchika4 
I am able to reproduce the error now. I was losing the `?unitylibs=stage` query parameter before, because of the microsoft login redirect.

robert-bogos (approved):

PR #166, opened 2024-11-26
MWPW-162758-Add cookie check for upload events

Ruchika4:
- Add cookie check for upload events before redirect

Resolves: [MWPW-162758](https://jira.corp.adobe.com/browse/MWPW-162758)

Test URLs:

Before: https://stage--cc--adobecom.hlx.page/acrobat/online/sign-pdf?unitylibs=stage
After: https://stage--cc--adobecom.hlx.page/acrobat/online/sign-pdf?unitylibs=analytics-cookie

Note: As the cookie is set only for adobe.com domain test by accessing above url on stage.adobe.com and overriding the action-binder.js with the code form this feature branch

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -168,7 +168,7 @@ export default class ActionBinder {
         {
           detail: {
             code,
-            message: `${message || 'Unable to process the request'}`,
+            message: `${message}`,
             status,
             info,
             accountType: this.getAccountType(),
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -278,15 +278,11 @@ export default class ActionBinder {
         this.block.dispatchEvent(new CustomEvent(unityConfig.trackAnalyticsEvent, { detail: { event: 'redirect to product' } }));
         await this.waitForCookie(2000);
         this.updateProgressBar(this.splashScreenEl, 100);
-        if (this.checkCookie()) {
-          window.location.href = response.url;
-        }
-        else {
+        if (!this.checkCookie()) {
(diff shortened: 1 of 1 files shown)


Ruchika4:
@TsayAdobe @joaquinrivero , could you please check once the wait time we have added and confirm if it should be ok. We are waiting for 2000 ms for cookie. We are also waiting for 500 ms before redirecting to product page when cookie not found so that the message can be logged to lana.

unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -251,11 +272,21 @@ export default class ActionBinder {
...
+        this.updateProgressBar(this.splashScreenEl, 100);
+        if (this.checkCookie()) {
+          window.location.href = response.url;
+        }
+        else {
+          await this.dispatchErrorToast('verb_cookie_not_set', 200, "Not all cookies found, redirecting anyway", true);
+          const sleep = (delay) => new Promise((resolve) => setTimeout(resolve, delay));
+          await sleep(500);
+          window.location.href = response.url;
+        }
```
TsayAdobe, line 289: It (L281-L289) looks like equivalent to the code below
```JS
        if (!this.checkCookie()) {
          await this.dispatchErrorToast('verb_cookie_not_set', 200, "Not all cookies found, redirecting anyway", true);
          await new Promise(r => setTimeout(r, 500));
        }
        window.location.href = response.url;
```

TsayAdobe (approved):

unitylibs/core/workflow/workflow-acrobat/action-binder.js
Ruchika4: Thanks @TsayAdobe I have incorporated suggested change.

sanjayms01 (approved):

Ruchika4:
Merging to this stage as this PR cannot be tested on hlx.page url without browser override.

cc @sanjayms01

PR #182, opened 2025-01-02
MWPW-163779-Support multi file upload for compress pdf

Ruchika4:
Handling multi file use case for Compress PDF

Resolves: [MWPW-163779](https://jira.corp.adobe.com/browse/MWPW-163779)

Test URLs:

Before: https://stage--dc--adobecom.hlx.page/acrobat/online/test/compress-pdf?unitylibs=stage
After: https://stage--dc--adobecom.hlx.page/acrobat/online/test/compress-pdf?unitylibs=multifile

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -85,6 +85,10 @@ export default class ActionBinder {
           this.promiseStack = [];
           await this.fillsign(files, eventName);
           break;
+        case value.actionType === 'compress':
+          this.promiseStack = [];
+          await this.compress(files, eventName);
+          break;    
         case value.actionType === 'continueInApp':
           this.LOADER_LIMIT = 100;
           await this.continueInApp();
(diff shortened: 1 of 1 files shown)


unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -85,6 +85,10 @@ export default class ActionBinder {
           this.promiseStack = [];
           await this.fillsign(files, eventName);
           break;
+        case value.actionType === 'compress':
```
TsayAdobe, line 88: Is it `compress` or `compress-pdf`? When do we use `compress` and when `compress-pdf`?

joaquinrivero (approved):

unitylibs/core/workflow/workflow-acrobat/action-binder.js
sanjayms01: This particular check is internal to Unity. We had called the action as "compress" but maybe @Ruchika4 we can make this change to be consistent in the target-config? What do you think?

Ruchika4: it is `compress` as we are matching the action type here https://github.com/adobecom/unity/blob/stage/unitylibs/core/workflow/workflow-acrobat/target-config.json#L68
`compress-pdf` is only for the verb-widget which is defined here https://github.com/adobecom/unity/blob/stage/unitylibs/core/workflow/workflow-acrobat/target-config.json#L39

sanjayms01 (approved):

unitylibs/core/workflow/workflow-acrobat/action-binder.js
Ruchika4: I think we can keep it `compress` only. @TsayAdobe please let us know if you see any concern.

TsayAdobe (approved):

PR #192, opened 2025-01-10
[MWPW-163793] Dispatch analytics events for Compress PDF

robert-bogos:
**Description:**
This PR enhances the analytics events dispatched to provide relevant data on the consumer side.  

**Resolves:** [MWPW-163793](https://jira.corp.adobe.com/browse/MWPW-163793)

**Test URLs:**
Before: https://stage--dc--adobecom.hlx.page/acrobat/online/test/compress-pdf?unitylibs=stage
After: https://stage--dc--adobecom.hlx.page/acrobat/online/test/compress-pdf?unitylibs=mwpw-163792-analytics

URL for PSI check: https://mwpw-163792-analytics--unity--adobecom.hlx.page/drafts/ruchika/remove-background?martech=off

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -622,7 +622,7 @@ export default class ActionBinder {
     this.block.dispatchEvent(
       new CustomEvent(
         unityConfig.trackAnalyticsEvent,
-        { detail: { event: eventName, data: { count: fileCount } } },
+        { detail: { event: eventName } },
       ),
     );
     this.block.dispatchEvent(new CustomEvent(unityConfig.trackAnalyticsEvent, { detail: { event: 'multifile', data: fileCount } }));
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -533,6 +533,7 @@ export default class ActionBinder {
       ),
     );
     let assetData = null;
+    const startTime = new Date();
     try {
       await this.showSplashScreen(true);
       const blobData = await this.getBlobData(file);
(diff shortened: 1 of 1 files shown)


unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -611,14 +612,17 @@ export default class ActionBinder {
...
+    this.block.dispatchEvent(new CustomEvent(unityConfig.trackAnalyticsEvent, { detail: { event: 'uploadTime', data: { uploadTime } } }));
     this.block.dispatchEvent(new CustomEvent(unityConfig.trackAnalyticsEvent, { detail: { event: 'uploaded' } }));
   }
 
   async multiFileUpload(fileCount, eventName) {
     this.block.dispatchEvent(
       new CustomEvent(
         unityConfig.trackAnalyticsEvent,
-        { detail: { event: eventName } },
+        { detail: { event: eventName, data: {count: fileCount}} },
```
sanjayms01, line 625: `fileCount` data is already being passed below when dispatching the `multifile` analytics event so believe this can be removed from here. This was tracked as a separate event based on analytics team wanting to track a new event for this. If this has since changed and you have some additional context for this, let's sync on this.

robert-bogos: I've added this to create some consistency with the same `'change'` [event that is sent for the single file upload](https://github.com/adobecom/unity/blob/stage/unitylibs/core/workflow/workflow-acrobat/action-binder.js#L524L534).
I can remove it if you don't think is necessary 🤔

Ruchika4: Hi @robert-bogos , as per requirement for multiple file upload we wanted a new analytics event for count of files hence I was sending it with the new event (`multifile`). Here is the ENB ticket https://jira.corp.adobe.com/browse/ENB-7550 Once this is done, Slytherin will have to listen to `multifile` event and pass the correct event name to analytics with number of files selected. 

Tagging @sanjayms01 also to confirm on above understanding.

cc @TsayAdobe  @joaquinrivero

unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -636,6 +641,7 @@ export default class ActionBinder {
     setTimeout(() => {
       this.updateProgressBar(this.splashScreenEl, 95);
       if (!this.redirectUrl) return;
+      this.block.dispatchEvent(new CustomEvent(unityConfig.trackAnalyticsEvent, { detail: { event: 'redirectComplete', data: this.redirectUrl } }));
```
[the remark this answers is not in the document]
sanjayms01: This has been confirmed that it is handled by DC product team.

robert-bogos:
Thanks @sanjayms01 ! I'm closing it

PR #200, opened 2025-01-23
[MWPW-160436]:-  Text-To-Image Prompt bar/Floating Prompt Bar

hnvadobe:
* Text to image Express changes.
* Specific
* Features or fixes

Resolves: https://jira.corp.adobe.com/browse/MWPW-160436

**Test URLs:**
- Before:  https://stage--cc--adobecom.hlx.page/drafts/hnv/text2image-test-page?unitylibs=stage&unityversion=v2&martech=off
- After: https://stage--cc--adobecom.hlx.page/drafts/hnv/text2image-test-page?unitylibs=t2i-latest&unityversion=v2&martech=off

### DIFF
--- a/unitylibs/core/workflow/workflow-ai/widget.css
+++ b/unitylibs/core/workflow/workflow-ai/widget.css
@@ -86,22 +86,27 @@
   display: flex;
   justify-content: center;
   visibility: visible;
-  padding: 10px 0;
 }
 .aside.text2image.unity-enabled .interactive-area .ex-unity-wrap.sticky .ex-unity-widget {
-  transform: translateY(-10px);
-  transition: transform 0.5s cubic-bezier(0.34, 1.56, 0.64, 1);
+  
 }
 .aside.text2image.unity-enabled .interactive-area .ex-unity-wrap.sticky.hidden .ex-unity-widget {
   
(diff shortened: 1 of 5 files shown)


sanjayms01:
@hnvadobe When I initially created this workflow, I had labelled it as `workflow-genai`, but I think we can now shorten it to just `workflow-ai`. Could we please make this update?

unitylibs/core/workflow/workflow-ai/widget.css
```diff
@@ -0,0 +1,420 @@
+.aside.text2image.unity-enabled {
```
sanjayms01, line 1: @hnvadobe I apologize if I have caused any confusion regarding `text2image` class. I spoke with @Ruchika4 on this and wanted to clarify that we should not introduce a new class (`text2image`) and rather leverage existing `unity-enabled` class from code for widget styles. As we want to generalize this workflow for more AI verbs in the future, we also should standardize the CSS accordingly without making it verb-specific. Please go ahead and remove `text2image` class.

hnvadobe:
I have raised a separate PR (https://github.com/adobecom/unity/pull/210) to resolve the conflict.

PR #277, opened 2025-02-25
[MWPW-168306, MWPW-168382] Enhancing logging of acrobat workflow errors

sanjayms01:
- Catch the IMS account type error and show a generic error toast
- Enhance logging of all API requests for acrobat workflow

Resolves: [MWPW-168306](https://jira.corp.adobe.com/browse/MWPW-168306)
Resolves: [MWPW-168382](https://jira.corp.adobe.com/browse/MWPW-168382)

**Test URLs:**
- After: https://stage--dc--adobecom.hlx.page/acrobat/online/compress-pdf?unitylibs=acrobat-logs

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -71,16 +71,18 @@ class ServiceHandler {
             if (resJson.reason?.includes(errorMessage)) error.message = errorMessage;
           });
         }
+        if (!error.message) error.message = `Error fetching from service. URL: ${url}, Options: ${JSON.stringify(options)}`;
         error.status = response.status;
         throw error;
       }
       if (contentLength === '0') return {};
       return response.json();
-    } catch (error) {
-      if (error.name === 'TimeoutError' || error.name === 'AbortError') {
-        error.status = 504;
+    } catch (e) {
+      if (['TimeoutError', 'AbortError'].includes(e.name)) {
+        e.status = 504;
+        e.message = `Request timed out. URL: ${url}, Options: ${JSON.stringify(options)}`;
       }
-      throw error;
+      throw e;
     }
   }
 
@@ -132,6 +134,7 @@ export default class ActionBinder {
     this.uploadHandler = null;
     this.splashScreenEl = null;
     this.promiseStack = [];
+    this.accountType = '';
     this.redirectUrl = '';
     this.redirectWithoutUpload = false;
     this.LOADER_DELAY = 800;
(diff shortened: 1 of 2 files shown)


unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -177,7 +179,11 @@ export default class ActionBinder {
   }
 
   getAccountType() {
-    return window.adobeIMS?.getAccountType?.() || '';
+    try {
+      return window.adobeIMS.getAccountType();
+    } catch (e) {
+      return '';
+    }
```
sharmeeatadobe, line 186: Since we are introducing a try-catch block, do you want to handle or log the error accordingly? Otherwise try-catch may not be needed

unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -177,7 +179,11 @@ export default class ActionBinder {
   }
 
   getAccountType() {
-    return window.adobeIMS?.getAccountType?.() || '';
+    try {
+      return window.adobeIMS.getAccountType();
```
aishwaryamathuria, line 183:
```suggestion
      return window.adobeIMS?.getAccountType ? window.adobeIMS.getAccountType() : '';
```

aishwaryamathuria: We can remove the try catch with this and avoid any error as well

sanjayms01: When testing this out, I am still seeing an exception raised in console when invoking this function with your suggested changes. Hence, I have kept the try-catch here and logging the error before performing the upload. Are you seeing otherwise?

unitylibs/core/workflow/workflow-acrobat/action-binder.js
sanjayms01: Yes, I am logging the error in the `handleSingleFileUpload` and `handleMutliFileUpload` functions.

sanjayms01: Good point, I have updated to log `e` within the try-catch.

sharmeeatadobe: After adding the change to log `e`, there will be two error toasts now. One at line 189 where you logged now, and then again inside the functions `handleSingleFileUpload` and `handleMutliFileUpload` since `getAccountType` will return an empty value. Is that fine?

aishwaryamathuria (approved):

unitylibs/core/workflow/workflow-acrobat/action-binder.js
aishwaryamathuria: for logged in users it should have a value and logged out users it should be guest. On which domain are you getting an error? probably the IMS guest might not be enabled?
Try and catch works as well though. We can verify the error as IMS should always return a value here.

unitylibs/core/workflow/workflow-acrobat/action-binder.js
sanjayms01: Updated to keep within function `getAccountType` function itself.

sharmeeatadobe (approved):

PR #287, opened 2025-03-03
MWPW-169062 | Adding animation to widget button 

aishwaryamathuria:
Adding animation to widget button

Resolves: [MWPW-169062](https://jira.corp.adobe.com/browse/MWPW-169062)

**Test URLs:**
- Before: https://main--unity--adobecom.aem.live/drafts/mathuria/remove-background?martech=off
- After: https://MWPW-169062--unity--adobecom.aem.live/drafts/mathuria/remove-background?martech=off

PR #276, opened 2025-02-25
Load hlx milo libs for hlx and aem for aem

aishwaryamathuria:
Load hlx milo libs for hlx and aem for aem

**Test URLs:**
- Before: https://main--unity--adobecom.aem.page/?martech=off
- After: https://hlxandaem--unity--adobecom.aem.page/?martech=off

PR #288, opened 2025-03-03
[Release] Stage to Main

Ruchika4:
* https://github.com/adobecom/unity/pull/287
* https://github.com/adobecom/unity/pull/276

**Test URLs:**
Before: https://main--unity--adobecom.aem.live/drafts/mathuria/remove-background?martech=off
After: https://main-hlx5--unity--adobecom.aem.live/drafts/mathuria/remove-background?martech=off

### DIFF
--- a/unitylibs/core/workflow/workflow-photoshop/widget.js
+++ b/unitylibs/core/workflow/workflow-photoshop/widget.js
@@ -43,7 +43,16 @@ export default class UnityWidget {
     if (continueInApp) this.addFeatureButtons('continue-in-app', continueInApp.closest('li'), unityaa, unityoa, '');
     const { decorateDefaultLinkAnalytics } = await import(`${getLibs()}/martech/attributes.js`);
     decorateDefaultLinkAnalytics(iWidget);
+    this.addBtnAnimation();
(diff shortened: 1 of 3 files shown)


rgclayton (approved):

unitylibs/core/workflow/workflow-photoshop/widget.js
```diff
@@ -43,9 +43,18 @@ export default class UnityWidget {
     if (continueInApp) this.addFeatureButtons('continue-in-app', continueInApp.closest('li'), unityaa, unityoa, '');
     const { decorateDefaultLinkAnalytics } = await import(`${getLibs()}/martech/attributes.js`);
     decorateDefaultLinkAnalytics(iWidget);
+    this.addBtnAnimation();
     return this.actionMap;
   }
 
+  addBtnAnimation() {
```
rgclayton, line 50: Should this be called before the initWidget function?

Ruchika4: As both the functions are in same class the function definition will be available when called within another method even though the function definition is later.

PR #306, opened 2025-03-12
MWPW-168306 Retrieving user state from Server-Timing header

sanjayms01:
- Performance Optimization: Checking user state based on response headers instead of waiting for IMS
- Removing unnecessary generic errors

Resolves: [MWPW-168306](https://jira.corp.adobe.com/browse/MWPW-168306)

**Test URLs:**
- After: https://stage--dc--adobecom.hlx.page/acrobat/online/compress-pdf?unitylibs=MWPW-168306-header


sanjayms01:
@sigadamvenkata @skumar09 When testing on my feature branch, I was able to verify on my end that these changes did not break any functionality. This fix would be best to verify on stage because the `Server-Timing` header is not present on the .hlx page but would appreciate your confirmation.

robert-bogos (approved):

PR #333, opened 2025-04-08
MWPW-170939 Passing user attempts to Connector API

sanjayms01:
- Parse user attempts count from localStorage and pass data to product via redirect URL

Resolves: [MWPW-170939](https://jira.corp.adobe.com/browse/MWPW-170939)

**Test URLs:**
- After: https://stage--dc--adobecom.aem.page/acrobat/online/compress-pdf?unitylibs=MWPW-170939

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -360,7 +360,18 @@ export default class ActionBinder {
   }
 
   async handleRedirect(cOpts) {
-    cOpts.payload.newUser = localStorage.getItem('unity.user') ? false : true;
+    try {
+      cOpts.payload.newUser = !localStorage.getItem('unity.user');
+      const numAttempts = parseInt(localStorage.getItem(`${this.workflowCfg.enabledFeatures[0]}_attempts`), 10) || 0;
+      const trialMapping = {
+        0: '1st',
+        1: '2nd',
+      };
+      cOpts.payload.attempts = trialMapping[numAttempts] || '2+';
+    } catch (e) {
+      cOpts.payload.newUser = true;
+      cOpts.payload.attempts = '1st';
+    }
     await this.getRedirectUrl(cOpts);
     if (!this.redirectUrl) return false;
     this.dispatchAnalyticsEvent('redirectUrl', this.redirectUrl);


unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -361,6 +361,12 @@ export default class ActionBinder {
 
   async handleRedirect(cOpts) {
     cOpts.payload.newUser = localStorage.getItem('unity.user') ? false : true;
+    const numAttempts = localStorage.getItem(`${this.workflowCfg.enabledFeatures[0]}_attempts`) || 0;
+    const attemptsAnalyticsMap = {
+      0: '1st',
+      1: '2nd',
+    };
+    cOpts.payload.attempts = attemptsAnalyticsMap[numAttempts] || '2+';
```
Ruchika4, line 369:
```suggestion
    const attemptLabel = numAttempts < 2 ? (numAttempts === 0 ? '1st' : '2nd') : '2+';
    cOpts.payload.attempts = attemptLabel;
```

Ruchika4 (changes requested):

unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -361,6 +361,12 @@ export default class ActionBinder {
 
   async handleRedirect(cOpts) {
     cOpts.payload.newUser = localStorage.getItem('unity.user') ? false : true;
+    const numAttempts = localStorage.getItem(`${this.workflowCfg.enabledFeatures[0]}_attempts`) || 0;
```
rohitsahu, line 364:
```suggestion
    const numAttempts = parseInt(localStorage.getItem(`${this.workflowCfg.enabledFeatures[0]}_attempts`)) || 0;
```

rohitsahu: Is there a possibility localStorage is not available? Incognito browsing or maybe user has explicitly disabled from browser settings. Should we have a null check or try catch block to be safer?

rohitsahu (approved):
minor comments. RLGTM.

unitylibs/core/workflow/workflow-acrobat/action-binder.js
sanjayms01: Hi @Ruchika4, thanks for the suggestion. I originally did think through having a nested ternary but also noticed that this approach is not encouraged by ESLint. Looking into it further, I notice that the performance impact between both approaches is rather negligible (between increase in file size & execution time) so instead vouched for the approach that was more maintainable and readable. Please let me know your thoughts or if you have any concerns.

unitylibs/core/workflow/workflow-acrobat/action-binder.js
sanjayms01: Good callout, I've added a try catch block to be safe.

unitylibs/core/workflow/workflow-acrobat/action-binder.js
Ruchika4: I don't have any major concerns. However I did felt this approach would be more cleaner and less lines specially in long run as we onboard more verbs. But I see the ESLint complaining about it. I'll go ahead and approve it.

Ruchika4 (approved):

rohitsahu (approved):

sigadamvenkata:
This PR has been tested with branch for Fill&Sign, compress verbs. now the parameter 'attempts' passed to product page with values (1st, 2ed, 2+)

below are sequence of URLs for positive test:
https://stage.acrobat.adobe.com/us/en/compress-pdf?x_api_client_id=unity&x_api_client_location=compress-pdf&user=frictionless_new_user&attempts=1st

https://stage.acrobat.adobe.com/us/en/compress-pdf?x_api_client_id=unity&x_api_client_location=compress-pdf&user=frictionless_return_user&attempts=2ed

https://stage.acrobat.adobe.com/us/en/compress-pdf?x_api_client_id=unity&x_api_client_location=compress-pdf&user=frictionless_return_user&attempts=2+

Also verified attempts sequence by uploading some error file by that product will throw an error. in that case 'attempts' param not being considered which is working as expected.

sample screen shot:
{15484DC5-F222-4BE0-8658-C2F753474F56}

PR #336, opened 2025-04-09
MWPW-170939 Passing user attempts to Connector API

sanjayms01:
- Parse user attempts count from localStorage and pass data to product via redirect URL [Stage Test]

Resolves: [MWPW-170939](https://jira.corp.adobe.com/browse/MWPW-170939)

**Test URLs:**
- After: https://stage--dc--adobecom.aem.page/acrobat/online/compress-pdf?unitylibs=MWPW-170939-stage-test


rohitsahu (approved):
lgtm

PR #344, opened 2025-04-10
[MWPW-170874] AWS Security credentials are being logged on splunk

rohitsahu:
Do not log request headers, it can contain access token. 
Do not log storage URL, it contains the AWS credentials

Resolves: [MWPW-170874](https://jira.corp.adobe.com/browse/MWPW-170874)

**Test URLs:**
- Before: https://stage--dc--adobecom.aem.page/acrobat/online/compress-pdf?unitylibs=stage&martech=off
- After: https://stage--dc--adobecom.aem.page/acrobat/online/compress-pdf?unitylibs=rosahu-MWPW-170874&martech=off


rohitsahu:
For now I have removed the request headers from being logged. Let's discuss, if we need to log any of those and the best way and format to log.

PR #364, opened 2025-04-18
[MWPW-171722]Code to handle different text for desktop in transition screen

Ruchika4:
* Code to handle different text for desktop in transition screen

Resolves: [MWPW-171722](https://jira.corp.adobe.com/browse/MWPW-171722)

**Test URLs:**
Test pages to make sure nothing breaks on live experience
- Before: https://stage--cc--adobecom.hlx.page/products/photoshop/online?martech=off&unitylibs=stage
- After: https://stage--cc--adobecom.hlx.page/products/photoshop/online?martech=off&unitylibs=MWPW-171722

Test page to make sure if 2 different text is authored on transition screen fragment before CC phase code goes live doesn't break anything on live experience:
- Before: https://stage--cc--adobecom.hlx.page/products/photoshop/online?martech=off&unitylibs=stage
- After: https://stage--cc--adobecom.hlx.page/products/photoshop/online?martech=off&unitylibs=splash-test

splash-test feature branch points to a transition fragment which has the updated authoring with 2 different text. As part of this PR we need to test and make sure that there is no change in existing experience in both the scenarios when 2 different texts are authored or when only 1 test is authored in transition screen fragment.

### DIFF
--- a/unitylibs/scripts/transition-screen.js
+++ b/unitylibs/scripts/transition-screen.js
@@ -3,16 +3,17 @@ import {
   localizeLink,
   loadImg,
   loadArea,
-} from '../scripts/utils.js';
+} from './utils.js';
 
 export default class TransitionScreen {
-  constructor(splashScreenEl, initActionListeners, loaderLimit, workflowCfg) {
+  constructor(splashScreenEl, initActionListeners, loaderLimit, workflowCfg, isDesktop = false) {
     this.splashScreenEl = splashScreenEl;
     this.initActionListeners = initActionListeners;
     this.LOADER_LIMIT = loaderLimit;
     this.workflowCfg = workflowCfg;
     this.LOADER_DELAY = 800;
     this.LOADER_INCREMENT = 30;
+    this.isDesktop = isDesktop;
   }
 
   updateProgressBar(layer, percentage) {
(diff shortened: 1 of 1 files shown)


unitylibs/scripts/transition-screen.js
```diff
@@ -126,11 +127,25 @@ export default class TransitionScreen {
     this.splashScreenEl.parentElement?.classList.add('hide-splash-overflow');
   }
 
+  updateCopyForDevice() {
+    const headingElements = this.splashScreenEl.querySelectorAll('h1, h2, h3, h4, h5, h6');
+    const mobileHeading = headingElements[1];
+    const desktopHeading = headingElements[2];
+    if (this.isDesktop && desktopHeading) {
+      if (mobileHeading) mobileHeading.style.display = 'none';
```
rohitsahu, line 135: Please check if just hiding the element using display none also accounts for the accessibility. Otherwise you can set 'aria-hidden' attribute as true/false depending on display none/block

rohitsahu (approved):
lgtm

unitylibs/scripts/transition-screen.js
rohitsahu: Yes, that's correct. I got confused with visibility: none.

unitylibs/scripts/transition-screen.js
```diff
@@ -126,11 +127,25 @@ export default class TransitionScreen {
     this.splashScreenEl.parentElement?.classList.add('hide-splash-overflow');
   }
 
+  updateCopyForDevice() {
+    const headingElements = this.splashScreenEl.querySelectorAll('h1, h2, h3, h4, h5, h6');
+    const mobileHeading = headingElements[1];
+    const desktopHeading = headingElements[2];
+    if (this.isDesktop && desktopHeading) {
```
rohitsahu, line 134: nit: suggested refactoring:
```  const hasDesktopHeading = !!desktopHeading;
  if (mobileHeading) {
    mobileHeading.style.display = (this.isDesktop && hasDesktopHeading) ? 'none' : 'block';
  }
  if (desktopHeading) {
    desktopHeading.style.display = (this.isDesktop && hasDesktopHeading) ? 'block' : 'none';
  }```

File: unitylibs/scripts/transition-screen.js [hunk 3]
```diff
@@ -126,11 +127,25 @@ export default class TransitionScreen {
     this.splashScreenEl.parentElement?.classList.add('hide-splash-overflow');
   }
 
+  updateCopyForDevice() {
+    const headingElements = this.splashScreenEl.querySelectorAll('h1, h2, h3, h4, h5, h6');
```
rohitsahu, line 131: querySelectorAll is an expensive op. Will it help if we cache the mobileHeading and desktopHeading  elements in the constructor.
Also, just wanted to confirm that picking  mobileHeading and desktopHeading is based on the assumption that the block always includes at least two headings (for mobileHeading) and three headings (for desktopHeading) and the order of the headings is kind of fixed, 1st maybe title or something, 2nd is for mobile and 3rd for thr desktop.

rgclayton (approved):

unitylibs/scripts/transition-screen.js
rohitsahu: Makes sense.

PR #366, opened 2025-04-21
Add upload success query parameter

rgclayton:
Adds `UTS_Uploaded` query param to the redirect url.
* If upload IS successful, `UTS_Uploaded=${this.uploadTimestamp}` is added to the end of the URL with the timestamp originating from the time of file upload.

Example URL with new query param:
`https://stage.acrobat.adobe.com/us/en/compress-pdf?x_api_client_id=unity&x_api_client_location=compress-pdf&user=frictionless_return_user&attempts=2%2B#assets=urn%3Aaaid%3Asc%3AUS%3A3f9a1c20-e9aa-31d6-8fc0-70d431897d85|ic-year-end-check-in-guide.pdf|8982279|application%2Fpdf&UTS_Uploaded=1745446585408`

Resolves: [MWPW-171687](https://jira.corp.adobe.com/browse/MWPW-171687)

**Test URLs:**
- Before: https://stage--dc--adobecom.hlx.page/acrobat/online/compress-pdf?martech=off
- After: https://stage--dc--adobecom.hlx.page/acrobat/online/compress-pdf?unitylibs=upload-success-response&martech=off

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -186,7 +186,7 @@ static ERROR_MAP = {
     this.MULTI_FILE = false;
     this.applySignedInSettings();
     this.initActionListeners = this.initActionListeners.bind(this);
-    this.uploadSuccess = false;
+    this.uploadTimestamp = null;
   }
 
   isSignedOut() {
@@ -521,8 +521,7 @@ static ERROR_MAP = {
       if (this.multiFileFailure && this.redirectUrl.includes('#folder')) {
         window.location.href = `${this.redirectUrl}&feedback=${this.multiFileFailure}`;
       } else {
-        const redirectUrlUploadSuccess = `${this.redirectUrl}${this.redirectUrl.includes('?') ? '&' : '?'}uploadSuccess=${this.uploadSuccess}`;
-        window.location.href = redirectUrlUploadSuccess;
+        window.location.href = this.redirectWithoutUpload === false ? `${this.redirectUrl}${this.redirectUrl.includes('?') ? '&' : '?'}UTS_Uploaded=${this.uploadTimestamp}` : this.redirectUrl;
       }
     } catch (e) {
       await this.transitionScreen.showSplashScreen();
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -186,6 +186,7 @@ static ERROR_MAP = {
     this.MULTI_FILE = false;
     this.applySignedInSettings();
     this.initActionListeners = this.initActionListeners.bind(this);
+    this.uploadSuccess = false;
   }
 
   isSignedOut() {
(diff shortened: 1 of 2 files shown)


unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -519,7 +520,10 @@ static ERROR_MAP = {
       await this.delay(500);
       if (this.multiFileFailure && this.redirectUrl.includes('#folder')) {
         window.location.href = `${this.redirectUrl}&feedback=${this.multiFileFailure}`;
-      } else window.location.href = this.redirectUrl;
+      } else {
+        const redirectUrlUploadSuccess = `${this.redirectUrl}${this.redirectUrl.includes('?') ? '&' : '?'}uploadSuccess=${this.uploadSuccess}`;
```
Ruchika4, line 524: We already have a variable this.redirectWithoutUpload which will be true if actual upload has not happen. Can we use that variable? And append the query param only when this.redirectWithoutUpload = false?

Ruchika4 (changes requested):

unitylibs/core/workflow/workflow-acrobat/action-binder.js
peller: I made a suggestion on slack that we might want to consider passing the timestamp instead of a boolean, using this query param for both AA and perf reporting. What do you think?  If we do that, it might be useful to use the two `UTS_` values as keys

Ruchika4 (approved):

unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -519,7 +520,9 @@ static ERROR_MAP = {
       await this.delay(500);
       if (this.multiFileFailure && this.redirectUrl.includes('#folder')) {
         window.location.href = `${this.redirectUrl}&feedback=${this.multiFileFailure}`;
-      } else window.location.href = this.redirectUrl;
+      } else {
+        window.location.href = this.redirectWithoutUpload === false ? `${this.redirectUrl}${this.redirectUrl.includes('?') ? '&' : '?'}UTS_Uploaded=${this.uploadTimestamp}` : this.redirectUrl;
+      }
```
Ruchika4, line 525: Product is removing everything after # in the redirect url hence adding query param in the end will also remove it.
@peller could you confirm if we should add this query param before the #assets in the url or DC product code will still be able to read it?
Here is a sample url https://stage.acrobat.adobe.com/us/en/compress-pdf?x_api_client_id=unity&x_api_client_location=compress-pdf&user=frictionless_return_user&attempts=2%2B#assets=urn%3Aaaid%3Asc%3AUS%3A06c736a2-8153-3620-8fe3-f212cceb3433|ic-goals-check-in-self-reflection-examples.pdf|6943103|application%2Fpdf If we open this in a browser everything after #asset is removed.

cc @rgclayton @sanjayms01

peller: query parameters go before hash, according to the URL spec. 
we are choosing to remove both the hash value and this new UTS query string from our code, just so the user doesn't have to look at them

rgclayton:
Closing. 

New PR:
https://github.com/adobecom/unity/pull/390

PR #373, opened 2025-04-23
[MWPW-170497] Unity PS phase2

Ruchika4:
* Unity PS Phase 2

Resolves: 
[MWPW-170497](https://jira.corp.adobe.com/browse/MWPW-170497)
[MWPW-170503](https://jira.corp.adobe.com/browse/MWPW-170503)
[MWPW-171707](https://jira.corp.adobe.com/browse/MWPW-171707)

**Test URLs:**
- Before: https://stage--cc--adobecom.hlx.page/drafts/ruchika/unity/online?unitylibs=stage
- After: https://stage--cc--adobecom.hlx.page/drafts/ruchika/unity/online?unitylibs=unity-ps-phase2


unitylibs/scripts/device-detection.js
rohitsahu, line 1: File should be inside utils folder.
Suggested filename: device-utils.js

Ruchika4: Thanks for the review, @rohitsahu I will move it inside utils. However I would prefer to keep the file name as device-detection as that gives more clarity on purpose of the file. Please let me know if any concern.

unitylibs/utils/device-detection.js
```diff
@@ -0,0 +1,120 @@
+function isIPad(userAgent) {
+  const ua = userAgent || (typeof navigator !== 'undefined' ? navigator.userAgent : '');
```
rohitsahu, line 2: Let's extract out ```userAgent || (typeof navigator !== 'undefined' ? navigator.userAgent : ''```

unitylibs/core/workflow/workflow-upload/action-binder.js
```diff
@@ -53,6 +55,7 @@ class ServiceHandler {
       alertText.innerText = msg;
       errorToast.classList.add('show');
     });
+    window.lana?.log(`Message: ${msg}, Error: ${errorUpdate}`, lanaOptions);
```
robert-bogos, line 58: nit: we could remove the `const errorUpdate ` above and just do:
```suggestion
    window.lana?.log(`Message: ${msg}, Error: ${error || ''}`, lanaOptions);
```

arugupta1992 (approved):

unitylibs/scripts/splunk-analytics.js
```diff
@@ -0,0 +1,50 @@
...
+  try {
+    const eventDataPayload = createPayloadForSplunk({ ...metaData, eventName });
+    fetch(splunkEndpoint, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(eventDataPayload),
+    });
+  } catch (error) {
+    window.lana?.log(
+      `An error occurred while sending ${eventName} to splunk, metadata: ${metaData}, error: ${error}`,
```
robert-bogos, line 46: `metaData` seems to be an object, so this will log `[object Object]`.
We should do:
```suggestion
      `An error occurred while sending ${eventName} to splunk, metadata: ${JSON.stringify(metaData)}, error: ${error}`,
```

rgclayton (approved):

asonnalagi:
Verified the PR. Noticed 2 observations on prelude screen. Logged tickets for those.
[MWPW-172200](https://jira.corp.adobe.com/browse/MWPW-172200)
[MWPW-172198](https://jira.corp.adobe.com/browse/MWPW-172198)

PR #389, opened 2025-05-07
MWPW-172495 Removing chunk upload failures from generic error code bucket

sanjayms01:
* Chunk upload failures have started to report into AA as they have categorized as generic errors
* We had previously fixed this by segregating these errors under a different error code
* This fix is to re-bucket these chunk upload failures so they are not tracked in AA as they have been mistakenly spiking up error rates 

Resolves: [MWPW-172495](https://jira.corp.adobe.com/browse/MWPW-172495)

**Test URLs:**
- After: https://stage--dc--adobecom.aem.page/acrobat/online/compress-pdf?unitylibs=MWPW-172495

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/upload-handler.js
+++ b/unitylibs/core/workflow/workflow-acrobat/upload-handler.js
@@ -92,7 +92,7 @@ export default class UploadHandler {
       if (!response.ok) {
         const error = new Error(response.statusText || 'Upload request failed');
         error.status = response.status;
-        await this.actionBinder.dispatchErrorToast('verb_upload_error_generic', response.status, `Failed when uploading chunk to storage; ${response.statusText}, ${assetId}, ${blobData.size} bytes`, true, true, {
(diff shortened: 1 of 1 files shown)


unitylibs/core/workflow/workflow-acrobat/upload-handler.js
```diff
@@ -92,7 +92,7 @@ export default class UploadHandler {
       if (!response.ok) {
         const error = new Error(response.statusText || 'Upload request failed');
         error.status = response.status;
-        await this.actionBinder.dispatchErrorToast('verb_upload_error_generic', response.status, `Failed when uploading chunk to storage; ${response.statusText}, ${assetId}, ${blobData.size} bytes`, true, true, {
+        await this.actionBinder.dispatchErrorToast('verb_upload_error_chunk_upload', response.status, `Failed when uploading chunk to storage; ${response.statusText}, ${assetId}, ${blobData.size} bytes`, true, true, {
```
rohitsahu, line 95: @sanjayms01 instead of error, do you want to keep it as warn since chunk upload failure does not necessarily lead to file upload failure?

sanjayms01: Good point, thanks. Updated to warn.

rohitsahu (approved):

Ruchika4 (approved):

PR #399, opened 2025-05-12
[MWPW-172186] Fix accessibility issues on transition screen

Ruchika4:
* Fix accessibility issues on transition screen

Resolves: [MWPW-172186](https://jira.corp.adobe.com/browse/MWPW-172186)

**Test URLs:**
DC
- After: https://stage--dc--adobecom.hlx.page/acrobat/online/compress-pdf?unitylibs=mwpw-172186-phase1

CC
- Before: https://stage--cc--adobecom.hlx.page/products/photoshop/online?unitylibs=stage
- After: https://stage--cc--adobecom.hlx.page/products/photoshop/online?unitylibs=mwpw-172186-phase1

### DIFF
--- a/unitylibs/core/styles/splash-screen.css
+++ b/unitylibs/core/styles/splash-screen.css
@@ -15,6 +15,18 @@ body > .splash-loader {
   z-index: 3;
 }
 
+.splash-loader .sr-only {
+  position: absolute !important;
+  width: 1px !important;
+  height: 1px !important;
+  padding: 0 !important;
+  margin: -1px !important;
+  overflow: hidden !important;
+  clip: rect(0, 0, 0, 0) !important;
+  white-space: nowrap !important;
+  border: 0 !important;
+}
+
 .splash-loader .section {
   height: 100%;
   width: 100%;
--- a/unitylibs/scripts/transition-screen.js
+++ b/unitylibs/scripts/transition-screen.js
@@ -143,11 +143,14 @@ export default class TransitionScreen {
     const desktopHeading = this.headingElements[2];
     if (mobileHeading) {
       mobileHeading.style.display = (this.isDesktop && desktopHeading) ? 'none' : 'block';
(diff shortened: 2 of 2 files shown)


sanjayms01 (approved):

unitylibs/core/styles/splash-screen.css
```diff
@@ -15,6 +15,18 @@ body > .splash-loader {
...
+.splash-loader .sr-only {
+  position: absolute !important;
+  width: 1px !important;
+  height: 1px !important;
+  padding: 0 !important;
+  margin: -1px !important;
+  overflow: hidden !important;
+  clip: rect(0, 0, 0, 0) !important;
+  white-space: nowrap !important;
+  border: 0 !important;
```
robert-bogos, line 27: I'm a bit sceptic on the heavy use of `!important`. We don't have another way around it?

Ruchika4: Use of !important in this context is common for "screen reader only" (sr-only) utility classes. It ensures that these visually hidden styles cannot be easily overridden by other CSS rules.

rgclayton: Could you scope `.sr-only` to the element it's on to raise the specificity? Would that remove the need for `!important`?

rgclayton (approved):
Other than my one comment. Looks good.

asonnalagi:
verified in Window chrome and iPhone. looks good but noticing a slight delay in % reading.
https://stage--dc--adobecom.hlx.page/acrobat/online/compress-pdf?unitylibs=mwpw-172186-phase1

robert-bogos (approved):

unitylibs/core/styles/splash-screen.css
robert-bogos: Got it @Ruchika4 ! 
If you have the time to explore what @rgclayton said it would be great, if not, given the time sensitivity given by today's release, I'm good with the code as is.

PR #426, opened 2025-05-20
[MWPW-173667] Upload errors are also exceptions which could be coming before uplaoding

rohitsahu:
Remove generic errors with more specific ones. Catch failures at more granular level to categorise them into pre upload and upload errors. Need slytherin code changes to support these errors
Corresponding Slytherin PR: https://github.com/adobecom/dc/pull/1193 

Resolves: [MWPW-173667](https://jira.corp.adobe.com/browse/MWPW-173667)

**Test URLs:**
- Before: https://stage--dc--adobecom.hlx.page/acrobat/online/compress-pdf
- After: https://stage--dc--adobecom.hlx.page/acrobat/online/compress-pdf?unitylibs=rosahu-MWPW-173667

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -164,6 +164,7 @@ static ERROR_MAP = {
   'pre_upload_error_empty_verb_limits': -51,
   'pre_upload_error_renaming_file' : -52,
   'pre_upload_error_fetch_redirect_url': -53,
+  'pre_upload_error_create_asset': -54,
   'validation_error_validate_files': -100,
   'validation_error_unsupported_type': -101,
   'validation_error_empty_file': -102,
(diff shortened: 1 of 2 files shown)


robert-bogos (approved):

unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -164,6 +164,7 @@ static ERROR_MAP = {
   'pre_upload_error_empty_verb_limits': -51,
   'pre_upload_error_renaming_file' : -52,
   'pre_upload_error_fetch_redirect_url': -53,
+  'pre_upload_error_create_asset': -54,
```
sanjayms01, line 167: I have added an error for the access token which is -54, so there will be a conflict. I'm fine to switch to -55 or if you want to switch let me know. Either way works for me.

sanjayms01 (approved):

PR #433, opened 2025-05-21
MWPW-173354 Adding enhanced logging for access token errors

sanjayms01:
- Adding enhanced logging for access token errors
- Provides messages if failure to retrieve token, failure to refresh, etc.

Resolves: [MWPW-173354](https://jira.corp.adobe.com/browse/MWPW-173354)

**Test URLs:**
- After: https://stage--dc--adobecom.hlx.page/acrobat/online/compress-pdf?unitylibs=MWPW-173354-token-logs

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -211,6 +211,7 @@ export default class ActionBinder {
     this.transitionScreen = null;
     this.promiseStack = [];
     this.signedOut = undefined;
+    this.tokenError = null;
     this.redirectUrl = '';
     this.filesData = {};
     this.errorData = {};
@@ -229,8 +230,12 @@ export default class ActionBinder {
   }
 
   async isSignedOut() {
-    const isGuest = await isGuestUser();
-    this.signedOut = isGuest ?? undefined;
+    const result = await isGuestUser();
+    if (result.error) {
+      this.tokenError = result.error;
+      return;
+    }
+    this.signedOut = result.isGuest ?? undefined;
   }
 
   setIsUploading(isUploading) {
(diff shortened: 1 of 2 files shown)


unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -588,16 +593,19 @@ export default class ActionBinder {
...
-        code: 'pre_upload_error_fetching_access_token',
-        subCode: 401,
-        desc: 'IMS access token could not be fetched',
-      });
-      return;
+      if (this.tokenError) {
+        const errorDetails = JSON.stringify(this.tokenError, null, 2);
+        await this.dispatchErrorToast('pre_upload_error_fetching_access_token', 401, `Could not fetch access token; Error: ${errorDetails}`, false, true, {
+          code: 'pre_upload_error_fetching_access_token',
+          subCode: 401,
```
rohitsahu, line 604: Do we need to hardcode this subCode?
This gives impression that token call failed due to backend returning 401

sanjayms01: That's a good point. However, since we don't get network response status, what would you suggest here?

unitylibs/scripts/utils.js
```diff
@@ -40,29 +40,74 @@ export {
...
 }
 
 async function getImsToken() {
-  const accessToken = window.adobeIMS?.getAccessToken();
-  if (!accessToken || accessToken?.expire.valueOf() <= Date.now() + (5 * 60 * 1000)) {
-    const refreshToken = await getRefreshToken();
-    return refreshToken;
+  try {
+    const accessToken = window.adobeIMS?.getAccessToken();
+    if (!accessToken) {
```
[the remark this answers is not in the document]
sanjayms01: Oops this was a miss. I had it in previously but seems like it was removed. I will bring the logic back to just refresh the token in this case.

unitylibs/core/workflow/workflow-acrobat/action-binder.js
rohitsahu: We can simply remove it. If undefined, it won't be added to the event data on splunk

unitylibs/scripts/utils.js
```diff
@@ -40,29 +40,79 @@ export {
...
 async function getImsToken() {
-  const accessToken = window.adobeIMS?.getAccessToken();
-  if (!accessToken || accessToken?.expire.valueOf() <= Date.now() + (5 * 60 * 1000)) {
-    const refreshToken = await getRefreshToken();
-    return refreshToken;
+  const RETRY_WAIT = 2000;
+  try {
+    const accessToken = window.adobeIMS?.getAccessToken();
+    if (!accessToken || accessToken?.expire.valueOf() <= Date.now() + (5 * 60 * 1000)) {
+      const firstAttempt = await attemptTokenRefresh();
```
rohitsahu, line 68: nit: line 68-71 is duplicate, can be extracted out

rohitsahu (approved):
1 nit, RLGTM

Ruchika4 (approved):

unitylibs/scripts/utils.js
sanjayms01: Thanks, will pick this up in a following PR.

PR #458, opened 2025-05-29
MWPW-174131 : Unable to upload on stage A.com page if user is already signed in

arugupta1992:
!isPdf() was being incorrectly sent instead of isPdf. Fixed now
Testing Notes: 
1. Check file upload for verbs that support single file pdf-only --- signed in user
2. Check for verbs that support non pdf - sfu and mfu --- signed in user

Resolves: [MWPW-174131](https://jira.corp.adobe.com/browse/MWPW-174131)

**Test URLs:**
- Before: https://stage--dc--adobecom.hlx.page/acrobat/online/merge-pdf?unitylibs=stage
- After: https://stage--dc--adobecom.hlx.page/acrobat/online/merge-pdf?unitylibs=MWPW-174131


rohitsahu (approved):
lgtm

Ruchika4 (approved):

PR #487, opened 2025-06-18
[MWPW-173462] Loading screen: Text should not be marked as a heading

rohitsahu:
* Replacing "One moment as we take you to acrobat.adobe.com" h2 tag with p. Updated css so that the tag change does not affect the text css.

Resolves: [MWPW-173462](https://jira.corp.adobe.com/browse/MWPW-173462)

Testing Notes:
The transition screen UI, specially the font size, style, weight and line height of the texts should remain same with and without this PR.
We also need to verify the above for backward compatibility. This can be done for validating the US pages and ROW pages. Authoring changes for US pages are supposed to be completed within a day or 2, the ROW cahnges can take some time. 
Validation needs to be done for Acrobat as well as Photoshop.

**Test URLs:**
- After: https://stage--dc--adobecom.aem.page/acrobat/online/compress-pdf?unitylibs=MWPW-173462

### DIFF
--- a/unitylibs/scripts/transition-screen.js
+++ b/unitylibs/scripts/transition-screen.js
@@ -185,7 +185,7 @@ export default class TransitionScreen {
       textNodes.forEach((node) => { node.textContent = ''; });
       if (this.splashScreenEl.querySelector('.icon-progress-bar')) await this.handleSplashProgressBar();
       if (this.splashScreenEl.querySelector('a.con-button[href*="#_cancel"]')) this.handleOperationCancel();
-      this.headingElements = this.splashScreenEl.querySelectorAll('h1, h2, h3, h4, h5, h6');
+      this.headingElements = this.splashScreenEl.querySelectorAll('h1, h2, h3, h4, h5, h6, p');
       this.splashScreenEl.setAttribute('aria-label', this.headingElements[1].innerText);
       if (this.workflowCfg.productName.toLowerCase() === 'photoshop') this.updateCopyForDevice();
       this.splashScreenEl.classList.remove('decorate');
--- a/unitylibs/scripts/transition-screen.js
+++ b/unitylibs/scripts/transition-screen.js
@@ -165,8 +165,8 @@ export default class TransitionScreen {
   }
 
   updateCopyForDevice() {
-    const mobileHeading = this.headingElements[1];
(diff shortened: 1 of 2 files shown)


unitylibs/scripts/transition-screen.js
```diff
@@ -77,6 +77,15 @@ export default class TransitionScreen {
     const resp = await fetch(`${this.splashFragmentLink}.plain.html`);
     const html = await resp.text();
     const doc = new DOMParser().parseFromString(html, 'text/html');
+    const headingToReplace = doc.querySelector('#one-moment-as-we-take-you-to-acrobatadobecom');
```
Ruchika4, line 80: I believe this will not work for localized version of transition screen as the id will be updated (Localized)

Ruchika4 (changes requested):

sanjayms01:
@rohitsahu Once you have the code fixes in place, please work with Anupama to ensure this gets verified on feature branch when you're back online. We'll get the PR merged to stage during PST time tomorrow.

sanjayms01:
@rohitsahu In discussions with @Ruchika4 and @rgclayton, it was agreed that it made sense for us to keep the backwards compatibility changes and fix them to ensure they work for both Ps and Acrobat. The reason for this is we do not yet have confirmation that ROW content will be updated in time, so we will need to ensure we still go live with a backwards-compatible solution. Once content launches are completed, we can go ahead and clean up/remove these changes after.

We have pinged both Kevin and Jessica in respective alignment channels, so please look to them for further updates on content launch coordination. Thanks!

unitylibs/scripts/transition-screen.js
rohitsahu: Thanks @Ruchika4 and @sanjayms01 for the feedback.
This code was added as a backup solution in case we don't get the authoring changes in time. I have removed it now, let me know if you see feel we should still bring this back.
Replacing all headings after product text might not be ideal as well and assuming the transition screen authoring will take care of that, it should not be needed.
Just curious, do we need to localise the id as well @Ruchika4? The only scenario I am able to think is if accessiblity needs it but those are usually needed in drop-down or multi select elements.

sanjayms01 (approved):

Ruchika4 (approved):

PR #497, opened 2025-07-05
Generic method to parse the error object (inckuding nested error object) and flatten it to be logged in splunk incase of ims token failure

arugupta1992:
Resolves: [MWPW-174041](https://jira.corp.adobe.com/browse/MWPW-174041)

**Test URLs:**
Before: https://stage--dc--adobecom.aem.page/acrobat/online/compress-pdf?unitylibs=stage
After: https://stage--dc--adobecom.aem.page/acrobat/online/compress-pdf?unitylibs=imsEnahncements-2

### DIFF
--- a/unitylibs/scripts/utils.js
+++ b/unitylibs/scripts/utils.js
@@ -44,10 +44,7 @@ async function getRefreshToken() {
   } catch (e) {
     return {
       token: null,
-      error: {
-        message: `Token refresh failed: ${e.message}`,
-        type: 'refresh_error',
-      },
+      error: e
     };
   }
 }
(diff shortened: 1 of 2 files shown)


rohitsahu (approved):
lgtm

sanjayms01 (changes requested):
Could we possibly instead move the `flattenObject` function out of the utils file and into a separate file? I understand the initial reasoning for keeping it here, however as this method is only called if the access token is null or expired, I don't believe we should incur a performance hit across all workflows by adding to utils which is used across all of our workflows. We can instead dynamically load it as needed. Please let me know your thoughts.

sanjayms01:
@arugupta1992 Please include the ticket number in PR title as well.

unitylibs/scripts/utils.js
```diff
@@ -37,6 +37,8 @@ export {
   createTag, loadStyle, getConfig, loadLink, loadScript, localizeLink, loadArea,
 };
 
+
+
```
sanjayms01, line 41: nit: remove extra new lines

PR #508, opened 2025-07-18
[MWPW-174296] Fetch acrobat page variant ID from Target for A/B Testing with edit pdf

rohitsahu:
- Fetch target campaign for mboxid 'ACOM_UNITY_ACROBAT_EDITPDF_POC' for add-comment verb
- Using alloy sdk(_satellite.track) for fetch call 
- Added Unit Test
- Temporarily using dev endpoint of unity BE. To be reverted
- Target code is present in a separate file `experiment-provider.js` for optimisation

Resolves: [MWPW-174296](https://jira.corp.adobe.com/browse/MWPW-174296)

**Test URLs:**
- Before: https://stage--dc--adobecom.aem.page/acrobat/online/compress-pdf
- After: https://stage--dc--adobecom.aem.page/acrobat/online/compress-pdf?unitylibs=MWPW-17429

### DIFF
--- a/test/utils/experiment-provider.test.js
+++ b/test/utils/experiment-provider.test.js
@@ -1,12 +1,11 @@
+/* eslint-disable no-underscore-dangle */
 import { expect } from '@esm-bundle/chai';
 import getExperimentData from '../../unitylibs/utils/experiment-provider.js';
 
 describe('getExperimentData', () => {
   beforeEach(() => {
     // Mock window._satellite
-    window._satellite = {
-      track: () => {},
-    };
+    window._satellite = { track: () => {} };
   });
 
   afterEach(() => {
@@ -32,13 +31,7 @@ describe('getExperimentData', () => {
 
     window._satellite.track = (event, options) => {
       const mockResult = {
-        decisions: [{
-          items: [{
-            data: {
-              content: mockTargetData,
-            },
-          }],
-        }],
+        decisions: [{ items: [{ data: { content: mockTargetData } }] }],
         propositions: ['test-proposition'],
       };
       setTimeout(() => {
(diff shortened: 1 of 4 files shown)


unitylibs/scripts/utils.js
```diff
@@ -320,8 +320,8 @@ export const unityConfig = (() => {
       ...commoncfg,
     },
     stage: {
-      apiEndPoint: 'https://unity-stage.adobe.io/api/v1',
```
rohitsahu, line 323: To be reverted

Ruchika4 (approved):

unitylibs/utils/experiment-provider.js
```diff
@@ -0,0 +1,29 @@
+export default async function getExperimentData() {
+  const decisionScopes = ['ACOM_UNITY_ACROBAT_EDITPDF_POC'];
```
sanjayms01, line 2: I think this was briefly touched on before but wanted to ask again, could this be a value inputted via authoring? Possibly via a query param maybe if not within the block? This is currently very specific to Edit PDF, but want to ensure we have a scalable solution in mind for future tests.

Note: I am not requesting any changes to this current PR as we have a strict deadline for this test. Just wanted to call this out and ensure we have thoughts on how to improve this further after this initial test.

test/utils/experiment-provider.test.js
```diff
@@ -0,0 +1,140 @@
...
+  afterEach(() => {
+    delete window._satellite;
+  });
+
+  it('should reject when target fetch fails', async () => {
+    window._satellite.track = (event, options) => {
+      setTimeout(() => {
+        if (typeof options.done === 'function') options.done(null, new Error('Test error'));
+      }, 0);
+    };
```
sanjayms01, line 20: I see variations of this "duplicated" in other parts of this file. Could we somehow abstract it out into a separate function to remove some redundancies?

sanjayms01 (approved):

PR #515, opened 2025-07-25
MWPW-176896 - PS marquee widget mobile-image-bottom flicker

rgclayton:
**Context:**
The media breakpoint needs to match what's being set in Milo's `split marquee`. The Milo marquee uses [decorate.js](https://github.com/adobecom/milo/blob/stage/libs/utils/decorate.js#L409-L418) to manipulates the DOM order for `.foreground` and `.image .asset` divs based on viewport size. 

* Added media query to match what's in Milo's `decorate.js` when changing viewports to avoid the flicker at 600px when using variant `.mobile-image-bottom`.
* **Note:** Viewport needs to match the viewport set for marquee to avoid flickering.

Resolves: [MWPW-176896](https://jira.corp.adobe.com/browse/MWPW-176896)

**Test URLs:**
- Before: https://main--cc--adobecom.aem.page/creativecloud/photography/type/black-background-photography?unitylibs=ps-widget-mobile-image-bottom?martech=off
- After: https://main--cc--adobecom.aem.page/creativecloud/photography/type/black-background-photography?unitylibs=ps-widget-mobile-image-bottom&martech=off


sanjayms01 (approved):

robert-bogos (approved):

asonnalagi:
Verified with the libs given in the PR. All looks good. Test Results are in the task [MWPW-176896](https://jira.corp.adobe.com/browse/MWPW-176896)

PR #524, opened 2025-08-05
[MWPW-177381] : Switching summary generator from MFU to SFU

arugupta1992:
As per the updated requirements from the product, this PR consists config change to restrict summary-generator to SFU

Resolves: [MWPW-177381](https://jira.corp.adobe.com/browse/MWPW-177381)

**Test URLs:**
Before: https://stage--dc--adobecom.aem.live/acrobat/online/merge-pdf?unitylibs=stage
After: https://stage--dc--adobecom.aem.live/acrobat/online/merge-pdf?unitylibs=sumGenSFU


rohitsahu (approved):
lgtm

sanjayms01:
> hi @arugupta1992 please merge it to stage. we start full testing in staging. thanks

Hi @sigadamvenkata! We believe there's still some testing that can be done on the feature branch to verify some of the logic. We'd still like to push for doing basic verification on the feature branch where possible to reduce the likelihood of bugs appearing on stage, as that can have impacts to our consumers as well. In this particular case, I think it would be great to have your verification if now uploading multiple files should result in an error toast as expected for example. We understand that E2E testing can only occur on stage but if we can get your documented sign-off on those particular items that can be tested on the feature branch (vs. what cannot be) before we merge our PRs, we'd really appreciate that.

At the same time, we can understand if you're coming up against any timing constraints from your end so happy to help support where needed. I know you and your team have helped support feature branch testing so far across all of our workflows, so just want to recommend that we continue that where feasible.

@arugupta1992 If there's any other specific cases you know can be verified on the feature branch that would help Venkata to test, please go ahead and document them on this PR as well.

Thank you!

PR #526, opened 2025-08-06
[MWPW-177382] Increase max concurrent files count to 10 for high end devices

rohitsahu:
Increase batch size to 10 files for high end, 5 for mid and and 3 for low end machines.
Do not block a batch if task completed

Resolves: [MWPW-177382](https://jira.corp.adobe.com/browse/MWPW-177382)

**Test URLs:**
- Before: https://stage--dc--adobecom.aem.page/acrobat/online/compress-pdf
- After: https://stage--dc--adobecom.aem.page/acrobat/online/compress-pdf?unitylibs=MWPW-177382

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/upload-handler.js
+++ b/unitylibs/core/workflow/workflow-acrobat/upload-handler.js
@@ -11,9 +11,9 @@ export default class UploadHandler {
   }
 
   static UPLOAD_LIMITS = {
-    HIGH_END: { files: 3, chunks: 10 },
-    MID_RANGE: { files: 3, chunks: 10 },
-    LOW_END: { files: 2, chunks: 6 },
+    HIGH_END: { files: 10, chunks: 10 },
+    MID_RANGE: { files: 5, chunks: 10 },
+    LOW_END: { files: 3, chunks: 6 },
   };
 
   async createAsset(file, multifile = false, workflowId = null) {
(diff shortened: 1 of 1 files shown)


unitylibs/core/workflow/workflow-acrobat/upload-handler.js
```diff
@@ -11,9 +11,9 @@ export default class UploadHandler {
   }
 
   static UPLOAD_LIMITS = {
-    HIGH_END: { files: 3, chunks: 10 },
-    MID_RANGE: { files: 3, chunks: 10 },
-    LOW_END: { files: 2, chunks: 6 },
+    HIGH_END: { files: 10, chunks: 10 },
+    MID_RANGE: { files: 5, chunks: 10 },
+    LOW_END: { files: 3, chunks: 6 },
```
Ruchika4, line 16: Should we move these numbers to target-config.json instead of hardcoding in js file?

sanjayms01: Makes sense to me, let's keep the configurations in there instead.

Ruchika4: Thanks for addressing the review comment, Rohit. Looks good to me now.

Ruchika4 (approved):

PR #553, opened 2025-09-02
[Do not merge][MWPW-179252] Adding support for fixing Multi Region Unity Calls

vipu0303:
1. Added support for fetching the base url via pageConfig call
2. Making this call when the user initiates the workflow

Resolves: [MWPW-179252](https://jira.corp.adobe.com/browse/MWPW-179252)

**Test URLs:**
- Before: https://stage--dc--adobecom.aem.live/acrobat/online/compress-pdf?unitylibs=stage
- After: https://stage--dc--adobecom.aem.page/acrobat/online/compress-pdf?unitylibs=multiRegionPoc2

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -346,12 +346,74 @@ export default class ActionBinder {
   }
 
   getAcrobatApiConfig() {
+    console.log('getAcrobatApiConfig called');
     unityConfig.acrobatEndpoint = {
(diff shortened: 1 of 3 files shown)


unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -346,14 +346,76 @@ export default class ActionBinder {
   }
 
   getAcrobatApiConfig() {
+    console.log('getAcrobatApiConfig called');
```
arugupta1992, line 349: Remove console logs

PR #562, opened 2025-09-10
[MWPW-180374]Increase unit test coverage for firefly workflow

Ruchika4:
* Increase unit test coverage for firefly workflow to bring it above 90%

Resolves: [MWPW-180374](https://jira.corp.adobe.com/browse/MWPW-180374)

**Test URLs:**
NA

### DIFF
--- a/test/core/workflow/workflow.firefly.test.js
+++ b/test/core/workflow/workflow.firefly.test.js
@@ -31,7 +31,7 @@ describe('Firefly Workflow Tests', () => {
       '.inp-field': [{ actionType: 'autocomplete' }],
     };
 
-    sinon.stub(UnityWidget.prototype, 'loadPrompts').callsFake(function () {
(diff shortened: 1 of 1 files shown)


vipu0303:
Current stats after this PR:
image
Hi @Ruchika4 Should we also put the branch and function level checks in check-coverage.js (currently it checks for Line coverage only). For now we can start with 70% for branches and 85% for functions so that we don't fall below these.
We all can pick up improving these number as per bandwidth.

Ruchika4:
> Current stats after this PR: image Hi @Ruchika4 Should we also put the branch and function level checks in check-coverage.js (currently it checks for Line coverage only). For now we can start with 70% for branches and 85% for functions so that we don't fall below these. We all can pick up improving these number as per bandwidth.

Hi @vipu0303 yes, that sounds good. We can create a ticket for that in our backlog.

PR #737, opened 2026-04-08
[MWPW-190483] Fixing error observed in analytics + console errors on clicking cancel

arugupta1992:
1. Changed all TransitionScreen instantiation sites to be lazy — only creates a new TransitionScreen instance if one doesn't already exist (if (!this.transitionScreen)). Also fixes a bug where this.transitionScreen.splashScreenEl was used (referencing the old object) instead of this.splashScreenEl, and updates LOADER_LIMIT on the existing instance rather than re-constructing.
2. Improved error logging to clearly distinguish between different server failures. Currently there were wrong messaging logged for few errors. 
3. Fixed console error on clicking cancel. (EPV)

image

Resolves: [MWPW-190483](https://jira.corp.adobe.com/browse/MWPW-190483)

**Test URLs:**
- Before: https://main--cc--adobecom.aem.page/products/firefly/features/remove-object-from-photo?unitylibs=stage
- After: https://main--cc--adobecom.aem.page/products/firefly/features/remove-object-from-photo?unitylibs=followups

### DIFF
--- a/unitylibs/core/workflow/workflow-upload/action-binder.js
+++ b/unitylibs/core/workflow/workflow-upload/action-binder.js
@@ -121,8 +121,10 @@ export default class ActionBinder {
       this.uploadAbortController?.abort();
       sendAnalyticsEvent(new CustomEvent('Cancel|UnityWidget'));
       this.logAnalyticsinSplunk('Cancel|UnityWidget', { assetId: this.assetId });
-      const { default: TransitionScreen } = await import(`${getUnityLibs()}/scripts/transition-screen.js`);
-      this.transitionScreen = new TransitionScreen(this.transitionScreen.splashScreenEl, this.initActionListeners, this.LOADER_LIMIT, this.workflowCfg, this.desktop);
+      if (!this.transitionScreen) {
+        const { default: TransitionScreen } = await import(`${getUnityLibs()}/scripts/transition-screen.js`);
+        this.transitionScreen = new TransitionScreen(this.splashScreenEl, this.initActionListeners, this.LOADER_LIMIT, this.workflowCfg, this.desktop);
+      }
       await this.transitionScreen.showSplashScreen();
       const e = new Error('Operation termination requested.');
       const cancelPromise = Promise.reject(e);
@@ -241,8 +243,10 @@ export default class ActionBinder {
         window.lana?.log(`Message: Upload aborted, Error: ${e.message}`, this.lanaOptions);
         return false;
       }
-      const { default: TransitionScreen } = await import(`${getUnityLibs()}/scripts/transition-screen.js`);
-      this.transitionScreen = new TransitionScreen(this.transitionScreen.splashScreenEl, this.initActionListeners, this.LOADER_LIMIT, this.workflowCfg, this.desktop);
+      if (!this.transitionScreen) {
+        const { default: TransitionScreen } = await import(`${getUnityLibs()}/scripts/transition-screen.js`);
+        this.transitionScreen = new TransitionScreen(this.splashScreenEl, this.initActionListeners, this.LOADER_LIMIT, this.workflowCfg, this.desktop);
+      }
       await this.transitionScreen.showSplashScreen();
       this.serviceHandler.showErrorToast({ errorToastEl: this.errorToastEl, errorType: '.icon-error-request' }, e, this.lanaOptions);
       this.logAnalyticsinSplunk('Upload server error|UnityWidget', {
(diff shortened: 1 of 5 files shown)


unitylibs/utils/NetworkUtils.js
```diff
@@ -146,8 +146,10 @@ export default class NetworkUtils {
       }
       return await this.afterFetchFromService(url, response);
     } catch (error) {
-      if (error) error.message += `, Max retry delay exceeded for URL: ${url}`;
```
arugupta1992, line 149: **Reason** : The catch block unconditionally appends "Max retry delay exceeded" to the error message regardless of whether the polling ever started. If the request failed on the first attempt with no retries then the message "Max retry delay exceeded" is wrong. This makes the analytics log misleading. The fix distinguish between a polling timeout and a first-attempt fetch failure:

unitylibs/scripts/transition-screen.js
```diff
@@ -31,6 +31,7 @@ export default class TransitionScreen {
   }
 
   updateProgressBar(layer, percentage) {
+    if (!layer) return;
```
arugupta1992, line 34: This is to guard for the race scenarios where the upload process begins before splashfragment loads. 

**Trade-off:**
Now if the layer (i.e. splashscreen) is null, the workflow will progress to completion but user sees the page normally until the redirect fires as splashscreen was never loaded. This should be acceptable behaviour, since current behaviour blocks the user flow and throws error in this scenario.

unitylibs/core/workflow/workflow-upload/upload-handler.js
```diff
@@ -76,7 +78,7 @@ export default class UploadHandler {
     );
     const { failedChunks, attemptMap } = result;
     const totalChunks = Math.ceil(file.size / blockSize);
-    if (failedChunks.size > 0) {
+    if (failedChunks.size > 0 && !signal?.aborted) {
```
arugupta1992, line 81: This is to prevent sending this analytics if workflow was cancelled by user.

unitylibs/core/workflow/workflow-upload/action-binder.js
```diff
@@ -153,7 +167,17 @@ export default class ActionBinder {
       body: blobData,
       ...(signal && { signal }),
     };
-    const response = await fetch(storageUrl, uploadOptions);
+    let response;
+    try {
+      response = await fetch(storageUrl, uploadOptions);
+    } catch (e) {
+      if (e instanceof TypeError) {
```
arugupta1992, line 174: This is to account for correct error message if it is a network failure.

unitylibs/core/workflow/workflow-upload/action-binder.js
```diff
@@ -237,12 +262,14 @@ export default class ActionBinder {
...
       return true;
     } catch (e) {
-      if (this.uploadAbortController?.signal.aborted || e.name === 'AbortError') {
+      if (signal.aborted || e.name === 'AbortError') {
         window.lana?.log(`Message: Upload aborted, Error: ${e.message}`, this.lanaOptions);
         return false;
       }
-      const { default: TransitionScreen } = await import(`${getUnityLibs()}/scripts/transition-screen.js`);
-      this.transitionScreen = new TransitionScreen(this.transitionScreen.splashScreenEl, this.initActionListeners, this.LOADER_LIMIT, this.workflowCfg, this.desktop);
+      if (!this.transitionScreen) {
```
arugupta1992, line 269: Changed all TransitionScreen instantiation sites to be lazy — only creates a new TransitionScreen instance if one doesn't already exist (if (!this.transitionScreen)). Also fixes a bug where this.transitionScreen.splashScreenEl was used (referencing the old object) instead of this.splashScreenEl, and updates LOADER_LIMIT on the existing instance rather than re-constructing.

sanjayms01: Great catch!

unitylibs/core/workflow/workflow-upload/action-binder.js
```diff
@@ -321,9 +348,12 @@ export default class ActionBinder {
       payload,
     };
     try {
-      const { default: TransitionScreen } = await import(`${getUnityLibs()}/scripts/transition-screen.js`);
+      if (!this.transitionScreen) {
+        const { default: TransitionScreen } = await import(`${getUnityLibs()}/scripts/transition-screen.js`);
+        this.transitionScreen = new TransitionScreen(this.splashScreenEl, this.initActionListeners, this.LOADER_LIMIT, this.workflowCfg, this.desktop);
+      }
```
sanjayms01, line 354: Thoughts on breaking this out to a separate function to remove the redundant lines of code?

unitylibs/scripts/transition-screen.js
sanjayms01: This is really interesting. I would agree that instead of a blocking flow, would be better to still let user be redirected to product.

How common of a case has this been?

arugupta1992: This is third top error for most of the verbs. First being 401 which is user error so we can call this out the second most impacting error.

unitylibs/core/workflow/workflow-upload/action-binder.js
arugupta1992: Yeah, good point. We already have loadTransitionScreen function which does the same. I am not sure why that wasn't being used throughout. I can take this up in my next PR.

vipu0303 (approved):

asonnalagi:
Looks fine in branch URL. Test results are in the ticket :[MWPW-190483](https://jira.corp.adobe.com/browse/MWPW-190483)

sanjayms01 (approved):

PR #740, opened 2026-04-15
[MWPW-192702] Fixing issues observed from evaluating the errors in analytics dashboard. 

arugupta1992:
**Issue Fix Descriptions** : Explained in the comments
**Testing Notes:** Refer to [wiki](https://wiki.corp.adobe.com/spaces/adobedotcom/pages/3835093951/PR+Testing+Notes+%E2%80%94+Analytics+Upload+Reliability+Fixes)

Resolves: [MWPW-192702](https://jira.corp.adobe.com/browse/MWPW-192702)

**Test URLs:**
Before: https://main--cc--adobecom.aem.page/products/firefly/features/remove-object-from-photo?unitylibs=stage
After: https://main--cc--adobecom.aem.page/products/firefly/features/remove-object-from-photo?unitylibs=analyticsIssues

### DIFF
--- a/unitylibs/core/workflow/workflow-upload/action-binder.js
+++ b/unitylibs/core/workflow/workflow-upload/action-binder.js
@@ -186,30 +186,6 @@ export default class ActionBinder {
     }
   }
 
-  async scanImgForSafety(assetId, signal) {
-    if (signal?.aborted) {
-      const err = new Error('Operation aborted');
-      err.name = 'AbortError';
-      throw err;
-    }
-    const assetData = { assetId, targetProduct: this.workflowCfg.productName };
-    const optionsBody = { body: JSON.stringify(assetData), ...(signal && { signal }) };
-    const res = await this.serviceHandler.postCallToService(
-      this.apiConfig.endPoint.acmpCheck,
-      optionsBody,
-      {},
-      false,
-    );
-    if (res.status === 429 || (res.status >= 500 && res.status < 600)) {
-      if (signal?.aborted) {
-        const err = new Error('Operation aborted');
-        err.name = 'AbortError';
-        throw err;
-      }
-      setTimeout(() => { this.scanImgForSafety(assetId, signal); }, 1000);
-    }
-  }
-
   async uploadAsset(file) {
     const assetDetails = {
       targetProduct: this.workflowCfg.productName,
@@ -257,7 +233,7 @@ export default class ActionBinder {
         );
       } else {
         await this.uploadImgToUnity(href, id, file, file.type, signal);
-        await this.scanImgForSafety(this.assetId, signal);
+        await uploadHandler.scanImgForSafetyWithRetry(this.assetId, signal);
         this.logAnalyticsinSplunk('Upload Completed|UnityWidget', { assetId: this.assetId });
       }
       return true;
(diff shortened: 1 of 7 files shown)


unitylibs/core/workflow/workflow-upload/action-binder.js
```diff
@@ -257,7 +233,7 @@ export default class ActionBinder {
         );
       } else {
         await this.uploadImgToUnity(href, id, file, file.type, signal);
-        await this.scanImgForSafety(this.assetId, signal);
```
arugupta1992, line 260: The recursive call inside setTimeout inside scanImgForSafety is not awaited and its return value is discarded. This means Upload server error|UnityWidget is never logged for any HTTP-level failure from scanImgForSafety — and separately, the entire upload flow incorrectly reports success (Upload Completed|UnityWidget) even when the safety scan is failing in a detached infinite retry loop in the background.

unitylibs/utils/chunkingUtils.js
```diff
@@ -48,9 +48,11 @@ export async function createChunkUploadTasks(uploadUrls, file, blockSize, upload
         if (onChunkComplete) onChunkComplete(i, chunkNumber, result);
         return result;
       } catch (err) {
-        const chunkInfo = { chunkIndex: i, chunkNumber };
-        failedChunks.add(chunkInfo);
-        if (onChunkError) onChunkError(chunkInfo, err);
+        if (err.name !== 'AbortError') {
```
arugupta1992, line 51: A dedicated chunkAbortController is created per upload session. Its signal is merged with the user-cancel signal via AbortSignal.any and passed to all chunks. When any chunk reaches final failure, chunkAbortController.abort() is called before throwing — immediately cancelling all sibling in-flight HTTP requests. AbortErrors in createChunkUploadTasks are excluded from failedChunks so only genuine failures are counted.

unitylibs/utils/NetworkUtils.js
```diff
@@ -10,15 +10,18 @@ export default class NetworkUtils {
   async fetchWithTimeout(url, options = {}, timeoutMs = 60000) {
     const controller = new AbortController();
     const timeout = setTimeout(() => controller.abort(), timeoutMs);
-    const passedSignal = options.signal || controller.signal;
-    const mergedOptions = { ...options, signal: passedSignal };
+    const mergedSignal = options.signal
```
arugupta1992, line 13: Used options.signal || controller.signal — only one signal was ever active. If an external signal was provided, the 60s timeout controller was completely ignored, making timeouts silently broken.

FIx: AbortSignal.any([controller.signal, options.signal]) — both fire correctly. The abort-source check (if (options.signal?.aborted) throw e) re-throws external aborts as AbortError instead of incorrectly converting them to TimeoutError. This means sibling chunk cancellations now correctly propagate as AbortError and are handled silently.

unitylibs/utils/NetworkUtils.js
```diff
@@ -101,6 +104,15 @@ export default class NetworkUtils {
         const onSuccessWithAttempt = onSuccess ? (response) => onSuccess(response, attempt) : null;
         const onErrorWithAttempt = onError ? (error) => onError(error, attempt) : null;
         let response = await this.fetchFromService(url, options, onSuccessWithAttempt, onErrorWithAttempt);
+        if (!response) {
```
arugupta1992, line 107: With this if fetchFromService returns undefined, the loop delays and retries (or breaks on the final attempt) instead of crashing.

sanjayms01 (approved):

sigadamvenkata:
verified some flows as per given notes. 
when user in went offline during upload
asset api 500, finalize calls, chunk upload failures
when user cancel upload not log to report etc

screen shots are updated in story. will do through testing in stage from user prospective to make sure rest of positive flows are fine @arugupta1992 @sanjayms01

shasmish (approved):

PR #747, opened 2026-04-19
MWPW-191927 Implementing Config-Based A/B Testing

sanjayms01:
* Implementing fetch request to page config API post-LCP
* Scalable and backwards-compatible approach to fetch experiment data and leverage location header for successive API calls

Resolves: [MWPW-191927](https://jira.corp.adobe.com/browse/MWPW-191927)

**Test URLs:**
- Before: https://main--dc--adobecom.aem.page/acrobat/online/compress-pdf?unitylibs=stage
- After: https://main--dc--adobecom.aem.page/acrobat/online/compress-pdf?unitylibs=MWPW-191927

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -182,6 +182,7 @@ export default class ActionBinder {
     this.multiFileValidationFailure = false;
     this.initialize();
     this.experimentData = null;
+    this.experimentViaPageConfig = false;
     this.pageConfigLocation = null;
     this.pageConfigFetched = false;
   }
@@ -266,6 +267,7 @@ export default class ActionBinder {
       this.pageConfigLocation = pageConfig.location;
       if (pageConfig.config?.target?.enabled) {
         this.experimentData = await getExperimentData(pageConfig.config.target.decisionScopes);
+        this.experimentViaPageConfig = true;
       } else if (!this.experimentData && this.workflowCfg.targetCfg?.experimentationOn?.includes(verb)) {
         const { getDecisionScopesForVerb } = await import('../../../utils/experiment-provider.js');
         const decisionScopes = await getDecisionScopesForVerb(verb);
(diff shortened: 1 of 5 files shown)


unitylibs/scripts/utils.js
```diff
@@ -335,3 +337,14 @@ export function sendAnalyticsEvent(event) {
 export function getMatchedDomain(domainMap = {}, hostname = window.location.hostname) {
   return Object.keys(domainMap).find((domain) => domainMap[domain].some((pattern) => new RegExp(pattern).test(hostname)));
 }
+
+export async function fetchPageConfig({ product, verb }) {
+  try {
+    const url = `${unityConfig.pageConfigEndPoint}?product=${product}&action=${verb}`;
```
shasmish: We renamed the action query param to verb to keep it consistent with other implementation.

sanjayms01: Got it, have updated accordingly.

Ruchika4 (approved):

unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -263,6 +256,32 @@ export default class ActionBinder {
     await priorityLoad(parr);
   }
 
+  async ensurePageConfig() {
+    if (this.pageConfigFetched) return;
+    this.pageConfigFetched = true;
+    const verb = this.workflowCfg.enabledFeatures[0];
+    try {
+      const { fetchPageConfig } = await import('../../../scripts/utils.js');
+      const { default: getExperimentData } = await import('../../../utils/experiment-provider.js');
```
arugupta1992, line 265: Should this be imported when target is enabled?

shasmish (approved):

unitylibs/core/workflow/workflow-acrobat/action-binder.js
sanjayms01: Since it was required in both if and else conditions, I kept it outside for now. But I agree, it should only be required when target is enabled and will make this change once we can remove the backwards-compatibility.

PR #751, opened 2026-04-24
[MWPW-193670] Wave2.1 - Image to video changes

arugupta1992:
Testing Notes: 
1. Regress all the upload workflow acros products
2. Test the limits for each product

Resolves: [MWPW-193670](https://jira.corp.adobe.com/browse/MWPW-193670)

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/drafts/arugupta/image-to-video/doodlebug/image-to-video?unitylibs=stage
- After: https://main--da-cc--adobecom.aem.page/drafts/arugupta/image-to-video/doodlebug/image-to-video?unitylibs=wave2-video


unitylibs/core/widgets/prompt-bar-upload/prompt-bar-upload.js
```diff
@@ -0,0 +1,520 @@
...
+  return {
+    container, triggerBtn, nameContainer, menuIcon, list,
+  };
+}
+
+/**
+ * Wires open/close behaviour for a dropdown:
+ * - clicking the trigger toggles the menu (closing any other open menus)
+ * - clicking outside closes it
+ */
```
sanjayms01: Please remove comments

unitylibs/core/widgets/prompt-bar-upload/prompt-bar-upload.js
```diff
@@ -0,0 +1,520 @@
...
+    const { origin } = window.location;
+    const baseUrl = (origin.includes('.aem.') || origin.includes('.hlx.'))
+      ? `https://main--unity--adobecom.${origin.includes('.hlx.') ? 'hlx' : 'aem'}.live`
+      : origin;
+    const res = await fetch(`${baseUrl}/unity/configs/prompt/model-picker-video-sample.json`);
+    if (!res.ok) throw new Error('Failed to fetch video models.');
+    const json = await res.json();
+    this.models = json?.content?.data || [];
+    this.buildAspectRatioMap();
+  }
```
sanjayms01, line 206: This has become such a redundant function that I question if it can be extracted out of the widget files. What are your thoughts on this?

unitylibs/core/widgets/prompt-bar-upload/prompt-bar-upload.js
```diff
@@ -0,0 +1,520 @@
...
+        height: Number(heights[i]) || null,
+      }));
+    });
+  }
+
+  getAspectRatiosForModel(modelId) {
+    return this.aspectRatioMap[modelId] || [];
+  }
+
+
```
sanjayms01: nit: extra space can be removed

unitylibs/core/widgets/prompt-bar-upload/prompt-bar-upload.js
```diff
@@ -0,0 +1,520 @@
...
+    const rightSection = createTag('div', { class: 'pbu-right-section' });
+    rightSection.append(promptBarContainer);
+    return rightSection;
+  }
+
+  buildDropZone() {
+    const fileInput = createTag('input', {
+      type: 'file',
+      id: 'file-upload',
+      accept: 'image/jpeg,image/jpg,image/png,image/webp',
```
sanjayms01: One suggestion here is can we think of a scalable way to modify the `accept` arg to ensure greater flexibility for any custom file type requirements for this widget?

unitylibs/core/widgets/prompt-bar-upload/prompt-bar-upload.js
```diff
@@ -0,0 +1,520 @@
...
+      previewImg.src = '';
+    };
+
+    this.widgetWrap?.addEventListener('pbu-image-selected', (e) => showPreview(e.detail.file));
+    deleteBtn?.addEventListener('click', (e) => {
+      e.stopPropagation();
+      showDropZone();
+      this.widgetWrap?.dispatchEvent(new CustomEvent('pbu-delete-image'));
+    });
+  }
```
sanjayms01, line 570: Wondering if there’s a potential stale preview edge case here when users re-upload images quickly?

If a second image is selected before the first one fully loads, the previous onload handler could still fire, and the old object URL isn’t revoked until then. Could this lead to a brief flash of the old image or unnecessary memory usage?

Would it make sense to:

- Revoke any existing object URL before creating a new one
- Clear previewImg.src and onload before assigning the new image
- Track the current object URL to guard against race conditions with older onload callbacks

unitylibs/core/workflow/workflow-prompt-bar-upload/action-binder.js
```diff
@@ -0,0 +1,679 @@
...
+      const error = new Error('Operation failed');
+      error.status = response.status;
+      throw error;
+    }
+    if (!failOnError) return response;
+    return response.json();
+  }
+
+  showErrorToast(errorCallbackOptions, error, lanaOptions, errorType = 'server') {
+    const isLightroomServerError = this.workflowCfg.productName.toLowerCase() === 'lightroom' && errorType === 'server';
```
sanjayms01: Curious, why are we checking lightroom here specifically?

unitylibs/core/workflow/workflow-prompt-bar-upload/action-binder.js
```diff
@@ -0,0 +1,679 @@
...
+
+  getVerbFromDom() {
+    const verbEl = this.unityEl?.querySelector('[class*="icon-operation-"]');
+    if (verbEl) {
+      const verbClass = Array.from(verbEl.classList).find((cls) => cls.startsWith('icon-operation-'));
+      const fromDom = verbClass?.slice('icon-operation-'.length);
+      if (fromDom) return fromDom;
+    }
+    return this.workflowCfg?.enabledFeatures?.[0];
+  }
```
sanjayms01, line 139: I believe it was previously discussed that a better approach would be to extract verb from the block class token instead. What do you think?

unitylibs/core/workflow/workflow-prompt-bar-upload/action-binder.js
```diff
@@ -0,0 +1,679 @@
...
+        decorateDefaultLinkAnalytics(errholder);
+        mount.append(errholder);
+      });
+      return this.toastCanvasAreas[0]?.querySelector('.pbu-main .alert-holder')
+        || this.toastCanvasAreas[0]?.querySelector('.alert-holder');
+    } catch (e) {
+      window.lana?.log(`Message: Error creating error toast, Error: ${e}`, this.lanaOptions);
+      return null;
+    }
+  }
```
sanjayms01, line 203: If overall DOM structure remains consistent across workflows, I wonder if this could be generalized later such that we don't need to rebuild and can instead pass custom CSS where needed.

unitylibs/core/workflow/workflow-prompt-bar-upload/action-binder.js
```diff
@@ -0,0 +1,679 @@
...
+      this.logAnalytics('Upload server error|UnityWidget', {
+        errorData: { code: 'error-request', subCode: `uploadAsset ${e.status}`, desc: e.message },
+        assetId: this.assetId,
+      });
+      return false;
+    }
+  }
+
+  validatePrompt(query) {
+    if (query.length > 750) {
```
sanjayms01: Instead of hardcoding 750 here, maybe could be a class variable or part of config for better maintainability?

unitylibs/core/workflow/workflow-prompt-bar-upload/target-config.json
```diff
@@ -0,0 +1,35 @@
+{
+  "_defaults": {
+    "renderWidget": true,
+    "connectorApiEndPoint": "https://unity.adobe.io/api/v1/asset/connector",
```
sanjayms01: Any particular reason why this need to exist within config here?

unitylibs/core/workflow/workflow.js
```diff
@@ -34,6 +34,10 @@ class WfInitiator {
         `${widgetBase}/prompt-bar-style/prompt-bar-style.js`,
         `${widgetBase}/prompt-bar-style/prompt-bar-style.css`,
       ],
+      'prompt-bar-upload': [
+        `${widgetBase}/prompt-bar-upload/prompt-bar-upload.js`,
+        `${widgetBase}/prompt-bar-upload/prompt-bar-upload.css`,
+      ],
```
sanjayms01: Let's clean up this function so that we can dynamically call the relevant js and CSS files based on the widget token. We should avoid having to continuously add more widgets to the registry as that raises scalability concerns.

unitylibs/scripts/transition-screen.js
```diff
@@ -92,7 +83,7 @@ export default class TransitionScreen {
       return splashScreenConfig[`fragmentLink-${matchedDomain}`];
     }
     const productName = this.workflowCfg.productName.toLowerCase();
-    if (this.workflowCfg.name === 'workflow-upload') {
+    if (this.workflowCfg.name === 'workflow-upload' || this.workflowCfg.name === 'workflow-prompt-bar-upload') {
```
sanjayms01, line 86: Can you please remind me why we have specific checks for these workflows here?

unitylibs/core/workflow/workflow-prompt-bar-upload/action-binder.js
arugupta1992: We can discuss this further, but from an approach standpoint, I think both options are reasonable. My main concern is maintaining consistency across waves. At the moment, we already have two different class naming conventions:
  -   1. Wave 1: icon-verb-*
  -   2. Wave 1.5.1: icon-operation-*

If we switch approaches for this workflow, we’ll end up maintaining three different patterns for extracting the verb name since I don’t think it would be practical to request changes in pages from previous waves.

My concern is that, over time, this becomes harder to reason about and remember, especially when revisiting the codebase later. So IMO, we should continue with the current approach until we have an opportunity to make this consistent across waves. Wdyt?

unitylibs/core/widgets/prompt-bar-upload/prompt-bar-upload.js
arugupta1992: I did think about that but delayed to re architecture task 😅 
There are other functions as well like building of model picker, creating gen button etc. The ideal way should be to have one base widget class that has the common functions and then respective widgets inherit/import from it, but that may have performance concerns to be validated. Could we track this via ticket and keep it out of wave 2 scope?

unitylibs/scripts/transition-screen.js
arugupta1992: Not sure. This is added by Jingle 2 months back. To me it looks like unnecessary but worth checking with @JingleH  if there was any specific use case for this?

unitylibs/core/workflow/workflow-prompt-bar-upload/action-binder.js
arugupta1992: Good point. Added it to target config.

unitylibs/core/workflow/workflow-prompt-bar-upload/action-binder.js
arugupta1992: Can't agree more. There are lot of common functions that can be extracted out and reused across workflows.

unitylibs/core/workflow/workflow-prompt-bar-upload/action-binder.js
arugupta1992: copy paste error. Good catch.

unitylibs/core/widgets/prompt-bar-style/prompt-bar-style.css
```diff
@@ -1333,4 +1333,4 @@
     font-size: 64px;
     letter-spacing: -1.92px;
   }
-}
+}
```
vipu0303, line 1336: please revert this.

unitylibs/core/widgets/prompt-bar-upload/prompt-bar-upload.js
```diff
@@ -0,0 +1,516 @@
...
+
+  getSizeForAspectRatio(modelId, ratio) {
+    const sizes = this.sizeMap[modelId] || [];
+    const ratios = this.aspectRatioMap[modelId] || [];
+    const idx = ratios.indexOf(ratio);
+    return idx !== -1 ? sizes[idx] : null;
+  }
+
+  readFeatureFlags() {
+    this.showAspectRatio = !!this.el.querySelector('[class*="icon-show-aspect-ratio"]');
```
vipu0303, line 245: should we also check for the value as true or not for these keys?

unitylibs/core/widgets/prompt-bar-upload/prompt-bar-upload.js
```diff
@@ -0,0 +1,516 @@
...
+
+    const promptBarContainer = createTag('div', { class: 'pbu-prompt-bar-container' });
+    promptBarContainer.append(promptLabel, promptTextarea, controlsFooter);
+
+    const rightSection = createTag('div', { class: 'pbu-right-section' });
+    rightSection.append(promptBarContainer);
+    return rightSection;
+  }
+
+  buildDropZone() {
```
vipu0303, line 454: please check with Jordan once if the complete marquee should act as a dropzone for drag and drop, as this was a requirement in earlier upload-marquee workflows.

unitylibs/core/widgets/prompt-bar-upload/prompt-bar-upload.js
```diff
@@ -0,0 +1,516 @@
...
+    const deleteBtn = createTag('button', { type: 'button', class: 'pbu-delete-btn', 'aria-label': 'Remove image' });
+    deleteBtn.innerHTML = svgIcon('#unity-trash-icon');
+    const uploadSpinner = createTag('div', { class: 'pbu-spinner hidden', 'aria-label': 'Uploading', role: 'status' });
+    const preview = createTag('div', { class: 'pbu-preview hidden', 'aria-hidden': 'true' });
+    preview.append(previewImg, deleteBtn, uploadSpinner);
+
+    const wrap = createTag('div', { class: 'pbu-drop-zone-wrap' });
+    wrap.append(dropZone, selectSpinner, preview);
+    return {
+      wrap, dropZone, preview, previewImg, deleteBtn,
```
vipu0303: nit:

```suggestion
      return { wrap, dropZone, preview, previewImg, deleteBtn };
```

unitylibs/core/widgets/prompt-bar-upload/prompt-bar-upload.js
```diff
@@ -0,0 +1,516 @@
...
+
+    this.widgetWrap?.addEventListener('pbu-image-selected', (e) => showPreview(e.detail.file));
+    deleteBtn?.addEventListener('click', (e) => {
+      e.stopPropagation();
+      showDropZone();
+      this.widgetWrap?.dispatchEvent(new CustomEvent('pbu-delete-image'));
+    });
+  }
+
+  async initWidget() {
```
vipu0303, line 572: please check the left side of the upload marquee is taking more than 50% of the total width.

unitylibs/core/workflow/workflow-prompt-bar-upload/action-binder.js
```diff
@@ -0,0 +1,676 @@
...
+export default class ActionBinder {
+  constructor(unityEl, workflowCfg, block, canvasArea, actionMap = {}) {
+    this.unityEl = unityEl;
+    this.workflowCfg = workflowCfg;
+    this.block = block;
+    this.canvasArea = canvasArea;
+    this.actionMap = actionMap;
+    this.errorToastEl = null;
+    this.transitionScreen = null;
+    this.LOADER_LIMIT = 95;
```
vipu0303, line 91: why is this 95?

unitylibs/core/workflow/workflow-upload/action-binder.js
```diff
@@ -383,6 +413,21 @@ export default class ActionBinder {
...
+    const { getVideoDuration } = await import(`${getUnityLibs()}/utils/FileUtils.js`);
+    const duration = await getVideoDuration(file);
+    this.filesData = { ...this.filesData, duration };
+    if (this.limits.minDuration && duration < this.limits.minDuration) {
+      this.handleClientUploadError('.icon-error-videoduration', 'error-videoduration', 'Video is too short');
+      throw new Error('Video is too short');
+    }
+    if (this.limits.maxDuration && duration > this.limits.maxDuration) {
+      this.handleClientUploadError('.icon-error-videoduration', 'error-videoduration', 'Video is too long');
+      throw new Error('Video is too long');
```
vipu0303, line 426: do these need to be localized?

unitylibs/core/widgets/prompt-bar-upload/prompt-bar-upload.js
sanjayms01: This has been addressed.

unitylibs/core/widgets/prompt-bar-style/prompt-bar-style.css
arugupta1992: This looks like a git error. Nothing to revert here.

unitylibs/core/workflow/workflow-upload/action-binder.js
arugupta1992: This is for analytics.

unitylibs/core/widgets/prompt-bar-upload/prompt-bar-upload.js
arugupta1992: Not sure, can you elaborate the intention around that? By default we are taking it as false. If it needs to be added then it should be authored.

unitylibs/core/workflow/workflow-prompt-bar-upload/action-binder.js
arugupta1992: That's the max cap for progress bar avoid going beyond 100 without finishing the upload tasks.

unitylibs/core/widgets/prompt-bar-upload/prompt-bar-upload.js
arugupta1992: That's as per design

sanjayms01 (approved):

vipu0303 (approved):

asonnalagi:
Validation completed in the branch URL. Test results are in the task :[MWPW-193897](https://jira.corp.adobe.com/browse/MWPW-193897)

PR #752, opened 2026-04-25
Wave 2.1 - Upscale video changes

arugupta1992:
Testing Notes: 
1. Regress all the upload workflow acros products
2. Test the limits for each product

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/drafts/arugupta/image-to-video/doodlebug/upscale-video?unitylibs=stage
- After: https://main--da-cc--adobecom.aem.page/drafts/arugupta/image-to-video/doodlebug/upscale-video?unitylibs=upscaleVid

### DIFF
--- a/unitylibs/core/workflow/workflow-upload/action-binder.js
+++ b/unitylibs/core/workflow/workflow-upload/action-binder.js
@@ -86,7 +86,11 @@ export default class ActionBinder {
     this.LOADER_LIMIT = 95;
     const commonLimits = workflowCfg.targetCfg.limits || {};
     const productLimits = workflowCfg.targetCfg[`limits-${workflowCfg.productName.toLowerCase()}`] || {};
-    this.limits = { ...commonLimits, ...productLimits };
+    const featureLimits = (workflowCfg.enabledFeatures || []).reduce((acc, feature) => ({
+      ...acc,
+      ...(workflowCfg.targetCfg[`limits-${feature}`] || {}),
+    }), {});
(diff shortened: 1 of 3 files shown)


unitylibs/core/workflow/workflow-upload/action-binder.js
```diff
@@ -420,23 +445,30 @@ export default class ActionBinder {
...
       this.handleClientUploadError('.icon-error-filetype', 'error-filetype');
       return;
     }
     if (this.limits.maxFileSize < file.size) {
       this.handleClientUploadError('.icon-error-filesize', 'error-filesize');
       return;
     }
-    try { await this.checkImageDimensions(file); } catch (error) {
-      window.lana?.log(`Message: Error checking image dimensions, Error: ${error}`, this.lanaOptions);
+    const isVideo = file.type.startsWith('video/');
```
sanjayms01, line 463: Should we be solely relying on mime type here or should we check for extension?

arugupta1992: If the mimetype matches and file extension does not then the control will exit from line 457 above. If control reaches this check that means file extensions are valid, so I think we can keep mimetype here to just check if it's video or image. wdyt?

sanjayms01: That's a fair point, we can leave as is given we have proper check beforehand.

sanjayms01 (approved):

PR #778, opened 2026-05-12
[MWPW-194995]Fix delay with config-based A/B testing api initiation

Ruchika4:
* Fix delay with config-based A/B testing api initiation

Resolves: [MWPW-194995(https://jira.corp.adobe.com/browse/MWPW-194995)

**Test URLs:**
- Before: https://stage--da-dc--adobecom.aem.page/acrobat/online/compress-pdf?martech=off&unitylibs=stage
- After: https://stage--da-dc--adobecom.aem.page/acrobat/online/compress-pdf?martech=off&unitylibs=target-test

**Implementation details:**
`this.pageConfigPromise = null (line 189)` initializes the field so it's always defined on the instance.
`this.pageConfigPromise = this.ensurePageConfig() `(line 792, inside initActionListeners when b === this.block) fires the fetch the moment the verb-widget finishes rendering (for performance reasons), no await so it doesn't block listener setup. The promise is stored so it can be joined later.
`await (this.pageConfigPromise || this.ensurePageConfig()) `(line 580, in handleFileUpload) if the widget has rendered (normal path), we await the already in flight or resolved promise; if somehow pageConfigPromise is null (edge case: upload triggered before initActionListeners ran with the block), we fall back to calling it directly. The existing pageConfigFetched guard inside ensurePageConfig still prevents double execution.


sanjayms01 (approved):

shasmish (approved):

asonnalagi:
As we cannot validate all the flows in branch, giving partial sign off. After merging to stage will do complete regression. For now, all looks fine.Test results are in the JIRA: [MWPW-195078](https://jira.corp.adobe.com/browse/MWPW-195078)

PR #780, opened 2026-05-19
[MWPW-193816] skip file dimension check if not able to get width/height of image

vipu0303:
skip file dimension check if not able to get width/height of image
please check the slack thread for asset details:
https://adobe-mwp.slack.com/archives/C0AE5AP6Q9L/p1778625759895929
Resolves: [MWPW-193816](https://jira.corp.adobe.com/browse/MWPW-193816)

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/products/firefly/features/ai-image-expander
- After: https://main--da-cc--adobecom.aem.page/products/firefly/features/ai-image-expander?unitylibs=mwpw-193816

### DIFF
--- a/unitylibs/core/workflow/workflow-upload/action-binder.js
+++ b/unitylibs/core/workflow/workflow-upload/action-binder.js
@@ -401,10 +401,7 @@ export default class ActionBinder {
     let { width, height } = {};
     try {
       ({ width, height } = await getImageDimensions(file));
-    } catch (e) {
-      window.lana?.log(`Message: Skipping image dimension check (unreadable), Error: ${e}`, this.lanaOptions);
-      return null;
-    }
+    } catch (e) { return null; }
     const isMaxLimits = this.limits.maxWidth && this.limits.maxHeight;
     const isMinLimits = this.limits.minWidth && this.limits.minHeight;
     this.filesData = { ...this.filesData, width, height };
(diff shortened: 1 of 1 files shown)


unitylibs/core/workflow/workflow-upload/action-binder.js
```diff
@@ -398,7 +398,13 @@ export default class ActionBinder {
 
   async checkImageDimensions(file) {
     const { getImageDimensions } = await import(`${getUnityLibs()}/utils/FileUtils.js`);
-    const { width, height } = await getImageDimensions(file);
+    let { width, height } = {};
+    try {
+      ({ width, height } = await getImageDimensions(file));
+    } catch (e) {
+      window.lana?.log(`Message: Skipping image dimension check (unreadable), Error: ${e}`, this.lanaOptions);
```
arugupta1992: Please remove lana log for client side errors.

arugupta1992: Is this logic tested? getImageDimensions() throws an error if the filetupe cannot be determined. I believe if dimensions are not present it may not throw error. Can you check this.?

arugupta1992 (changes requested):

unitylibs/core/workflow/workflow-upload/action-binder.js
vipu0303: @arugupta1992 @sanjayms01 
even if the fileType is determined and if we are not able to extract the dimensions, it throws an exception which we are not catching and leading to block the user to redirect to FF without giving any proper error message.
We should anyways redirect the user to FF if we are not able to get the dimensions.

wrt to the lana logs, we are not sending any analytics in this case, so I believe we can have the lana log.

arugupta1992 (approved):

asonnalagi:
Validation done in the branch. All looks fine. Test results are in the task: [MWPW-195836](https://jira.corp.adobe.com/browse/MWPW-195836)

shasmish (approved):

PR #804, opened 2026-06-09
[MWPW-194359] Remove Background Unity changes 

vipu0303:
Epic - [Remove BG Epic](https://jira.corp.adobe.com/browse/MWPW-194359)
This PR include the changes related to the Remove Background story
Adds the workflow-inline-action experience for Firefly remove background on hero marquee pages. Introduces the inline-action widget (viewport-aware authoring, drop zone, inline loading progress, result view with re-upload/download, NBA cards, and Edit in Firefly).
Guest users complete remove background inline; signed-in users use the splash flow and redirect to Firefly. Upload reuses the shared UploadHandler with chunked upload support. Progress is driven manually at 60% / 95% / 100% for upload, remove background, and connector/complete. Download triggers a local save for the first user (with empty-blob validation) before falling back to the connector. Includes unit tests and workflow registration.

PRD: [Remove BG PRD](https://adobe-my.sharepoint.com/:w:/p/tek27852/IQDUYwduihAKS7vJX72OX53JAQpTO3m439UCGRDrB-VNWJQ?e=tSw0Nm)
[Figma Designs](https://www.figma.com/design/w2VbLPROC36BO0DA0wBgyY/Doodlebug?node-id=8895-70043&p=f&m=dev)
Unity FE-BE-FF [contract wiki](https://wiki.corp.adobe.com/pages/viewpage.action?pageId=3902725355&spaceKey=LEO&title=Doodlebug%2B-%2BRemove%2BBackground)

Resolves: [MWPW-194359](https://jira.corp.adobe.com/browse/MWPW-194359)

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/drafts/vipulg/doodlebug/rbg/vg-rbg
- After: https://main--da-cc--adobecom.aem.page/drafts/vipulg/doodlebug/rbg/vg-rbg?unitylibs=vg-rbg


test/core/workflow/mocks/inline-action-authored.js
```diff
@@ -0,0 +1,60 @@
+export default `<div>
```
arugupta1992, line 1: Where/Why do we need mocks? Is this for UTs only?

unitylibs/core/workflow/workflow-upload/upload-handler.js
```diff
@@ -63,10 +63,11 @@ export default class UploadHandler {
     return this.networkUtils.fetchFromServiceWithRetry(storageUrl, uploadOptions, retryConfig, onSuccess, onError);
   }
 
-  async uploadChunksToUnity(uploadUrls, file, blockSize, signal = null) {
+  async uploadChunksToUnity(uploadUrls, file, blockSize, signal = null, extraOptions = {}) {
```
arugupta1992: Can we rename extraOptions to onSuccess or successCallback?

unitylibs/core/workflow/workflow.js
```diff
@@ -120,6 +125,7 @@ class WfInitiator {
       actionBinderBlock,
       canvasAreaForBinder,
       this.actionMap,
+      unityWidgetObject,
```
arugupta1992, line 128: Why do we need to pass widget for this workflow? How is it working for other widgets?

unitylibs/core/workflow/workflow-inline-action/action-binder.js
```diff
@@ -0,0 +1,646 @@
...
+  }
+
+  async initActionListeners(b = this.block, actMap = this.actionMap) {
+    this.serviceHandler = new ServiceHandler(this.canvasArea, this.unityEl, this.getAdditionalHeaders.bind(this));
+    await this.initAnalytics();
+    if (this.workflowCfg.targetCfg.showSplashScreen) this.loadTransitionScreen();
+    Object.keys(actMap).forEach((key) => {
+      b.querySelectorAll(key).forEach((el) => {
+        const action = actMap[key];
+        if (el.classList.contains('drop-zone') || el.classList.contains('ia-dropzone')) return;
```
arugupta1992: What is the purpose of this condition? Why are we returning for this condition?

unitylibs/core/widgets/inline-action/inline-action.js
```diff
@@ -0,0 +1,393 @@
...
+      const sprite = createTag('div', { class: 'ia-sprite hide' });
+      sprite.innerHTML = this.spriteContent;
+      root.append(sprite);
+    }
+    root.append(left, right);
+    const unityRoot = insertInlineActionRoot(this.el, this, root);
+    setupInteractiveAreaDragAndDrop(unityRoot.querySelector('.interactive-area'), root, dropZone, fileInput);
+    this.widget = root;
+
+    return {
```
arugupta1992: This should be part of target-config.json.

unitylibs/core/widgets/inline-action/inline-action.js
```diff
@@ -0,0 +1,393 @@
...
+    const input = this.widget?.querySelector('.ia-file-input');
+    if (input) input.value = '';
+  }
+
+  openFilePicker() {
+    this.resetFileInput();
+    this.widget?.querySelector('.ia-file-input')?.click();
+  }
+
+  async initWidget() {
```
arugupta1992, line 441: async can be removed I believe. No async operation being done in this.

unitylibs/core/widgets/inline-action/inline-action.js
```diff
@@ -0,0 +1,393 @@
...
+    resultActions.append(reuploadBtn, downloadBtn);
+    checker.append(resultActions);
+    left.append(preview, buildGhostOverlay(), createTag('div', { class: 'ia-result' }, checker));
+
+    const progressHolder = TransitionScreen.createProgressBar();
+    const { dropZoneContainer, dropZone, fileInput } = buildDropZoneContainer(this.meta, progressHolder);
+    this.progressScreen = new TransitionScreen(progressHolder, () => {}, 100, this.workflowCfg);
+    this.progressScreen.progressText = this.meta.loadingText;
+
+    const grid = createTag('div', { class: 'ia-nba-grid' });
```
arugupta1992: Can this be extracted out as 'buildNbaGrid()?. Likewise consider the initWidget function to just orchestrate building of the final dom. Building of sections should be extracted out to respective functions like buildLeftPanel(), buildNbaGrid() etc.

unitylibs/core/widgets/inline-action/inline-action.js
```diff
@@ -0,0 +1,393 @@
...
+  if (skin === 'dark') el.classList.add('dark');
+  const root = createTag('div', { class: 'unity-inline-action unity-enabled' });
+  root.append(createTag('div', { class: `interactive-area ${skin}` }, widgetEl));
+  const holder = createTag('div', { class: 'ia-config-holder ia-sr-only', 'aria-hidden': 'true' });
+  while (el.firstChild) holder.append(el.firstChild);
+  el.append(holder);
+  el.classList.add('unity-inline-action-host');
+  if (el.parentNode) el.parentNode.insertBefore(root, el);
+  else el.append(root);
+  widgetInstance.promptBarExtendedRoot = root;
```
arugupta1992: promptBar?

unitylibs/core/widgets/inline-action/inline-action.js
```diff
@@ -0,0 +1,393 @@
...
+const svgUse = (id, className = '') => `<svg aria-hidden="true"${className ? ` class="${className}"` : ''}><use xlink:href="#${id}"></use></svg>`;
+
+function appendIconContent(el, { href, spriteId, size, picture = false }) {
+  if (href) {
+    const img = createTag('img', { src: href, alt: '', loading: 'lazy', ...(size && { width: size, height: size }) });
+    el.append(picture ? createTag('picture', {}, img) : img);
+  } else el.innerHTML = svgUse(spriteId);
+}
+
+function buildUploadActionButton(uploadIconHref, uploadLabel) {
```
arugupta1992: Can this be generic buildActionButton() and be used to build all the action buttons - download, upload, NBA arrow clicks, Reupload etc. This could probably take in parameters if action button is icon only or icon + text. This would centralise action button creation logic for all the instances. Thoughts?

unitylibs/core/widgets/inline-action/inline-action.js
```diff
@@ -0,0 +1,393 @@
+import { createTag, defineDeviceByScreenSize } from '../../../scripts/utils.js';
+import TransitionScreen from '../../../scripts/transition-screen.js';
```
arugupta1992: Please make this a dynamic import where required.

arugupta1992 (changes requested):

unitylibs/core/widgets/inline-action/inline-action.js
```diff
@@ -0,0 +1,393 @@
...
+    this.resetFileInput();
+    this.widget?.querySelector('.ia-file-input')?.click();
+  }
+
+  async initWidget() {
+    this.meta = parseInlineAuthoring(this.el);
+    const root = createTag('div', { class: 'ia-widget', 'data-state': InlineActionState.INITIAL });
+    const left = createTag('div', { class: 'ia-panel ia-panel-left' });
+    const preview = createTag('div', { class: 'ia-preview' }, createTag('img', { class: 'ia-preview-img', src: this.meta.heroSrc, alt: '' }));
+    const checker = createTag('div', { class: 'ia-checker' }, createTag('img', { class: 'ia-result-img', src: '', alt: 'Processed image' }));
```
arugupta1992: Consider lazy downloading these two as well. Also check for the NBA thumbnails.

unitylibs/core/widgets/inline-action/inline-action.js
```diff
@@ -0,0 +1,393 @@
...
+    if (input) input.value = '';
+  }
+
+  openFilePicker() {
+    this.resetFileInput();
+    this.widget?.querySelector('.ia-file-input')?.click();
+  }
+
+  async initWidget() {
+    this.meta = parseInlineAuthoring(this.el);
```
arugupta1992: nit: could we rename meta to something more meaningful like parsedData to understand it's coming via authoring?

unitylibs/core/workflow/workflow-inline-action/action-binder.js
```diff
@@ -0,0 +1,646 @@
...
+        await this.triggerDownload(this.resultUrl);
+        userCount = this.incrementUserCount();
+      } catch (e) {
+        // vipulg: consider here that download failed
+        this.serviceHandler.showErrorToast(this.uploadErrorOpts(), e, this.lanaOptions);
+        return;
+      }
+      // vipulg: consider here that download succeeded
+    }
+    await this.callConnector(await this.buildConnectorPayload({
```
arugupta1992, line 583: Rename to continueInApp to keep consistency with other workflows.

unitylibs/core/workflow/workflow.js
vipu0303: It is needed now bec the widget instance is now stateful as it no keeps the state and progress on the instance (setState, setProgress)

unitylibs/core/workflow/workflow-upload/upload-handler.js
vipu0303: Renamed the variable to "callbacks" it can contain both success and failure callbacks.

unitylibs/core/widgets/inline-action/inline-action.js
vipu0303: in workflow.js we await on this. So required.

unitylibs/core/widgets/inline-action/inline-action.js
vipu0303: We need to eager load only the image which gets loaded on page load. And thats the default image in ia-preview.

unitylibs/core/widgets/inline-action/inline-action.js
vipu0303: Not everything but some basic buttons can be clubbed. But at the same time it will add little logic as well. So there is a tradeoff, keeping things little short vs keeping UI creation simple. Will think more on this.

unitylibs/core/widgets/inline-action/inline-action.js
vipu0303: kept the name same as promptBarExtendedRoot, as it is being used in workflow.js.
Ideally we should rename this to something generic, maybe just extendedRoot. But then it would trigger update in both prompt-bar-style and prompt-bar-audio

unitylibs/core/workflow/workflow-inline-action/action-binder.js
vipu0303: updated. Moved inline-action drop-zone wiring to action-binder and actionMap to target-config.

unitylibs/core/widgets/inline-action/inline-action.js
vipu0303: moved inline-action drop-zone wiring to action-binder and actionMap to target-config

unitylibs/core/workflow/workflow-inline-action/action-binder.js
```diff
@@ -0,0 +1,756 @@
...
+        this.apiConfig.endPoint.assetUpload,
+        { body: JSON.stringify(assetDetails) },
+        this.uploadErrorOpts(),
+      );
+      if (signal.aborted) return false;
+      const { id, href, blocksize, uploadUrls } = resJson;
+      this.assetId = id;
+      this.logAnalyticsinSplunk('Asset Created|UnityWidget', { assetId: this.assetId });
+      const { default: UploadHandler } = await import(`${getUnityLibs()}/core/workflow/workflow-upload/upload-handler.js`);
+      const uploadHandler = new UploadHandler(this, this.serviceHandler);
```
sanjayms01, line 320: If there's functionality that can be shared between workflows, I believe it may be better for us to abstract this out to a common utils instead. What are your thoughts?

We don't need to make this change now, but just wanted to touch on possible improvements for the future.

unitylibs/core/workflow/workflow.js
```diff
@@ -273,6 +279,10 @@ class WfInitiator {
         sfList: new Set(['text-to-mage']),
         stList: new Set(['prompt', 'tip', 'legal', 'generate']),
       },
+      'workflow-inline-action': {
+        productName: product || 'Firefly',
```
sanjayms01, line 283: Is this future-proofing for potential other products we may need to support?

unitylibs/scripts/transition-screen.js
```diff
@@ -82,12 +82,15 @@ export default class TransitionScreen {
...
-      const { theme } = this.workflowCfg;
-      const themedKey = theme ? `fragmentLink-${productName}-${theme}` : null;
+    const productName = this.workflowCfg.productName?.toLowerCase();
+    if (productName) {
+      const themedKey = this.workflowCfg.theme === 'dark'
+        ? `fragmentLink-${productName}-dark`
+        : null;
       if (themedKey && splashScreenConfig[themedKey]) return splashScreenConfig[themedKey];
-      return splashScreenConfig[`fragmentLink-${productName}`];
+      if (splashScreenConfig[`fragmentLink-${productName}`]) {
```
sanjayms01, line 92: Might be good to have a null check for splashScreenConfig itself at the top of this function before just checking its entries.

sanjayms01 (approved):

unitylibs/core/workflow/workflow.js
vipu0303: Yes it should be I believe. We should be able to use this workflow in any other product. But yes once we do it, then only we will have to check e2e.

unitylibs/core/workflow/workflow-inline-action/action-binder.js
vipu0303: yeah.. This uploadAsset mentod is almost same wrt workflow-upload, prompt-bar-upload-widgets.
We may want to extract it out to utils maybe. Good callout.

arugupta1992 (approved):

asonnalagi:
Verified the flow from unity side in the branch. There are some major issues. No blocker or critical. Signing off to merge this to stage.

PR #813, opened 2026-06-15
[MWPW-198610] Mobile tab experience

vipu0303:
Ipad and tablets should follow mobile logic
 I've checked the Express RBG Experience; here's what I'm seeing: 

    On iOs/Safari first time user download scenario: For me it is only triggering the download and not redirecting to FF at all. I'm seeing the same thing unfortunately. 
    On iOS/chrome: Downloads the file and redirects to FF in same tab. This is correct, and what I see as well on Express (except they try to go to the app, we should go straight into Firefly)! 

Fallback for ios/safari if we are not able to achive download + redirect: 
1. "Only download and not redirect as Express does." Let's go with this fallback, because at least the user also sees "Open in Firefly" as a CTA on mobile, right? 
2. "Only redirect as user can still have a chance to download on product." Not Ideal User Journey as download is not as obvious in product

Resolves: [MWPW-198610](https://jira.corp.adobe.com/browse/MWPW-198610)

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.live/drafts/vipulg/doodlebug/rbg/vg-rbg?unitylibs=stage
- After: https://main--da-cc--adobecom.aem.live/drafts/vipulg/doodlebug/rbg/vg-rbg?unitylibs=iosFix

### DIFF
--- a/unitylibs/core/workflow/workflow-inline-action/action-binder.js
+++ b/unitylibs/core/workflow/workflow-inline-action/action-binder.js
@@ -15,8 +15,7 @@ import {
 } from '../../../scripts/utils.js';
 import { InlineActionState } from '../../widgets/inline-action/inline-action.js';
 import { INLINE_ACTION_EVENTS } from '../../../scripts/analytics.js';
-import isDesktop from '../../../utils/device-detection.js';
-import { getCgenQueryParams } from '../../../utils/cgen-utils.js';
+import isDesktop, { isIOS } from '../../../utils/device-detection.js';
 
 const DOWNLOAD_COUNT_KEY = 'inline-action-download-count';
 const WORKFLOW_NAME = 'inline-action';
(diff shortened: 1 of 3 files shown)


unitylibs/core/workflow/workflow-inline-action/action-binder.js
```diff
@@ -15,6 +15,7 @@ import {
 } from '../../../scripts/utils.js';
 import { InlineActionState } from '../../widgets/inline-action/inline-action.js';
 import { INLINE_ACTION_EVENTS } from '../../../scripts/analytics.js';
+import isDesktop, { isIOSSafari } from '../../../utils/device-detection.js';
```
sanjayms01: can we dynamically import these instead?

sanjayms01 (approved):

PR #814, opened 2026-06-15
[MWPW-197381]Onboard express for resume builder verb

Ruchika4:
* Onboard resume builder for express

Resolves: [MWPW-197381](https://jira.corp.adobe.com/browse/MWPW-197381)

**Test URLs:**
- Before: https://resume-builder--da-express-milo--adobecom.aem.page/drafts/methomas/express/create/resume?unitylibs=stage
- After: https://resume-builder--da-express-milo--adobecom.aem.page/drafts/methomas/express/create/resume?unitylibs=express-2

- New `verb-dropzone` block for Express pages — renders a dropzone area with upload button, drag-hint text, file limit text, and legal footer authored via the Unity block.

### Testing Notes:

Drag & Drop
- Drop listeners are attached to `document.body` (not scoped to a div), so dropping a file **anywhere on the page** triggers upload while the dropzone is in the viewport.
- Listeners are added/removed via `IntersectionObserver` — dropping a file when the dropzone is scrolled out of view should **not** trigger upload.

 File Input (Click to Upload)
- Clicking the upload button opens the file picker restricted to `image/jpeg, image/jpg, image/png, image/webp`.
- Selecting an unsupported file type should show a validation error toast.

Redirect Query Param (`x_api_client_location`)
- On redirect, the URL should contain `x_api_client_location` computed from the page path: strips `/express/` prefix, reverses path segments, joins with `_`.
- Example: `/express/create/resume` → `x_api_client_location=resume_create`
- Verify the correct value appears in the redirect URL for the page under test.
- Test page if created will need to be created inside a express folder

Transition Screen
- Upload should show a splash/transition screen (from `/express/unity/splashscreen` fragment, mounted on `body`).
- Verify it appears on file drop/select and clears on redirect.

User logged in status
- Test the flow for anonymous users. There is a discussion going on where we might need to support signed in users to upload files too. However this PR doesn't touch that yet. So scope for this PR is anonymous user and signed in user experience will be tested separately

Regression
- Please do a regression on an existing acrobat verb with unitylibs=express-2 to make sure everything works as expected


sanjayms01 (approved):

asonnalagi:
Validated from A.com side. All looks fine. Test results are in the task: [ MWPW-198704](https://jira.corp.adobe.com/browse/MWPW-198704)

arugupta1992 (approved):

PR #815, opened 2026-06-16
Changes required in remove background analytics to build splunk dashboard

arugupta1992:
1. Added parameter to classiffiy signedin vs signed out users
2. Added workflow parameter to all events
3. Synced the server error namin across events 
4. Wrapped connector cal in try catch to record it's failures and success

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/drafts/vipulg/doodlebug/rbg/vg-rbg?unitylibs=stage
- After: https://main--da-cc--adobecom.aem.page/drafts/vipulg/doodlebug/rbg/vg-rbg?unitylibs=rbg-dashboard


unitylibs/core/workflow/workflow-inline-action/action-binder.js
```diff
@@ -166,7 +195,12 @@ export default class ActionBinder {
   }
 
   logAnalyticsinSplunk(eventName, data) {
-    this.sendAnalyticsToSplunk?.(eventName, this.workflowCfg.productName, { ...data, workflow: WORKFLOW_NAME }, `${unityConfig.apiEndPoint}/log`);
+    this.sendAnalyticsToSplunk?.(
+      eventName,
+      this.workflowCfg.productName,
+      this.getAnalyticsMeta(data),
```
sanjayms01, line 201: `trackEvent` already calls this method and we're calling it again here?

arugupta1992: trackEvent calls logAnalyticsinSplunk. Here it is using sendAnalyticsToSplunk. Am i missing anything?

sanjayms01: Sorry I was referring to why `getAnalyticsMeta` is also being called here when it is called in `trackEvent` before we logAnalyticsInSplunk.

arugupta1992: Ohh, my bad. 
trackEvent is used when analytics are to be pushed to both AA and splunk. For building the splunk dashboard we add few more analytics which are pushed only to splunk, so at those points logAnalyticsinSplunk is called directly. For those cases getAnalyticsMeta is called here.

vipu0303 (approved):
lgtm

unitylibs/core/workflow/workflow-inline-action/action-binder.js
```diff
@@ -288,8 +323,7 @@ export default class ActionBinder {
       this.signedInFlowInProgress = false;
       this.assetId = null;
       this.resultAssetId = null;
-      sendAnalyticsEvent(new CustomEvent('Cancel|UnityWidget'));
-      this.logAnalyticsinSplunk('Cancel|UnityWidget', { assetId });
+      this.trackEvent('Cancel|UnityWidget', { assetId });
```
vipu0303, line 326: please define this string in analytics.js as we do for other similar strings.

unitylibs/core/workflow/workflow-inline-action/action-binder.js
```diff
@@ -560,8 +600,9 @@ export default class ActionBinder {
     this.resultUrl = null;
     this.resultBlob = null;
     this.filesData = { count: 1, size: file.size, type: file.type, name: file.name };
-    sendAnalyticsEvent(new CustomEvent('Uploading Started|UnityWidget'));
     const { isGuest } = await isGuestUser();
+    this.isGuestUser = isGuest;
+    this.trackEvent('Uploading Started|UnityWidget');
```
vipu0303, line 605: thanks for catching this

sanjayms01 (approved):

PR #826, opened 2026-06-19
[AEEE-35854][MWPW-199140] Firefly Logo on transition screen for express url

vipu0303:
Fix for correctly loading the media on the Transition screen.

What the old code did
The old code only ran replaceDotMedia for Acrobat, and resolved media paths using contentRoot from the current page's config.
Instead of using contentRoot, we construct the base URL from where the fragment is actually fetched and then resolve each ./media_* relative to that.
This is correct for every domain — Firefly, Acrobat, Express — because the media always lives next to the fragment, regardless of what contentRoot is or which page loads the fragment.

Resolves: 
[AEEE-35854](https://jira.corp.adobe.com/browse/AEEE-35854)
[MWPW-199140](https://jira.corp.adobe.com/browse/MWPW-199140)

For Validation use below URL:
https://www.adobe.com/express/feature/image/remove-background?at_preview_token=lLr7bJrDlMCVKw90nSqAxA&at_preview_index=1_2&at_preview_listed_activities_only=true&at_preview_evaluate_as_true_audience_ids=2515310
and replace the transition-screen.js file with the new file content.

**Test URLs:**
- Before: https://main--unity--adobecom.aem.page/?martech=off
- After: https://<branch>--unity--adobecom.aem.page/?martech=off

### DIFF
--- a/unitylibs/scripts/transition-screen.js
+++ b/unitylibs/scripts/transition-screen.js
@@ -96,19 +96,22 @@ export default class TransitionScreen {
     return splashScreenConfig.fragmentLink;
   }
 
-  replaceDotMedia(area = document) {
+  replaceDotMedia(area = document, fragmentBaseUrl = null) {
     const config = getConfig?.();
     const contentRoot = config?.contentRoot ?? '';
-    if (!contentRoot) return;
+    if (!fragmentBaseUrl && !contentRoot) return;
     const prefix = (config?.locale?.prefix ?? '').replace('/', '') ?? '';
     const currUrl = new URL(window.location.href);
     const pathSeg = currUrl.pathname.split('/').length;
-    if (area === document && ((prefix === '' && pathSeg >= 3) || (prefix !== '' && pathSeg >= 4))) return;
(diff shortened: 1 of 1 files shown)


sanjayms01:
@vipu0303 Thanks for pushing out this fix. Harry was actually able to resolve this issue on the T&O side, so we no longer need this fix on our end. We can go ahead and close this PR.

cc @asonnalagi @sigadamvenkata

vipu0303:
> @vipu0303 Thanks for pushing out this fix. Harry was actually able to resolve this issue on the T&O side, so we no longer need this fix on our end. We can go ahead and close this PR.
> 
> cc @asonnalagi @sigadamvenkata

Hi @sanjayms01  As discussed, for now since we have put the firefly asset in the express path, the media file path is getting resolved and working. But that is actually a workaround. We should actually resolve the path of the media used in a fragment, relative to the fragment path and not relative to the path where this fragment is being used.
So I still believe we should take this PR. Please review.

cc @arugupta1992 @Ruchika4 @asonnalagi @sigadamvenkata

PR #830, opened 2026-06-24
fix for correctly tracking splunk analytics

vipu0303:
fix: await connector success analytics before same-tab redirect

Connector success analytics were being lost on same-tab redirects because window.location.href unloaded the page before the in-flight Splunk fetch could complete.

Threaded the Promise return value up through sendAnalyticsToSplunk → logAnalyticsinSplunk → trackEvent → trackConnectorSuccess and awaited it in callConnector only on same-tab redirect paths — both openInSameTab and the popup-blocked window.open fallback. New-tab redirects do not await since the current page stays alive and the fetch completes naturally.

The change to analytics.js is additive — all other callers across workflows are already discarding the return value and are unaffected.

This PR needs validation across workflows and across different products.

Resolves: [MWPW-199504](https://jira.corp.adobe.com/browse/MWPW-199504)

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/products/firefly/features/remove-background
- After: https://main--da-cc--adobecom.aem.page/products/firefly/features/remove-background?unitylibs=splunkFix

### DIFF
--- a/unitylibs/core/workflow/workflow-inline-action/action-binder.js
+++ b/unitylibs/core/workflow/workflow-inline-action/action-binder.js
@@ -115,7 +115,7 @@ export default class ActionBinder {
     sendAnalyticsEvent(new CustomEvent(eventName, {
       detail: { workflow: metaData.workflow },
     }));
-    this.logAnalyticsinSplunk(eventName, data);
+    return this.logAnalyticsinSplunk(eventName, data);
   }
 
   trackRemoveBackgroundError(error) {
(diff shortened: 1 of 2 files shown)


Ruchika4 (approved):

sanjayms01 (approved):

sigadamvenkata:
to test the fix in safari MAC , we need to move the code to stage to remove the sidekick dependency. with this exception we can merge the PR to STAGE.

PR #831, opened 2026-06-24
MWPW-199518 Adding workaround for image aspect ratio distortion

sanjayms01:
* Fix distorted aspect ratio on Remove Background for iPhone portrait photos by correcting EXIF orientation before upload
* Add readExifOrientation to read the EXIF orientation tag from JPEG headers without loading the full image
* Add correctOrientation to redraw JPEG to canvas pre-upload, baking in correct pixel orientation for the RemoveBackground API
* Gate orientation correction to mobile only — desktop JPEGs are unaffected and incur no overhead
* Skip canvas re-encoding entirely if EXIF orientation is already correct (value 1), avoiding unnecessary processing for screenshots and pre-processed images

Resolves: [MWPW-199518](https://jira.corp.adobe.com/browse/MWPW-199518)

**Test URLs:**
- Before: https://stage--da-cc--adobecom.aem.live/products/firefly/features/remove-background?unitylibs=stage
- After: https://stage--da-cc--adobecom.aem.live/products/firefly/features/remove-background?unitylibs=MWPW-199518

**Testing Use Cases**
- Upload a recently taken iPhone portrait photo on adobe.com Remove Background page on iPhone Safari (iOS 18.x) and verify the result image is not distorted
- Upload the same photo on desktop browser and verify it processes correctly with no regression
- Upload a PNG and WebP file on iPhone Safari and verify they are unaffected by the change
- Upload a JPEG screenshot or downloaded image (no EXIF rotation) on iPhone Safari and verify the canvas re-encoding step is skipped (result should be identical to before)
- Upload a landscape iPhone photo and verify it processes correctly with no distortion

### DIFF
--- a/unitylibs/core/workflow/workflow-inline-action/action-binder.js
+++ b/unitylibs/core/workflow/workflow-inline-action/action-binder.js
@@ -22,6 +22,28 @@ const UPLOAD_ERROR_TYPE = '.icon-error-request';
 const PROGRESS = { UPLOAD_MAX: 60, REMOVE_BG: 95, COMPLETE: 100 };
 const getMount = (el) => el?.querySelector?.('.ia-widget') || el;
 
+function correctOrientation(file) {
+  if (!file.type.match(/image\/(jpeg|jpg)/i)) return Promise.resolve(file);
+  return new Promise((resolve) => {
+    const url = URL.createObjectURL(file);
+    const img = new Image();
+    img.onload = () => {
+      URL.revokeObjectURL(url);
+      const canvas = document.createElement('canvas');
+      canvas.width = img.naturalWidth;
+      canvas.height = img.naturalHeight;
+      canvas.getContext('2d').drawImage(img, 0, 0);
+      canvas.toBlob(
+        (blob) => resolve(blob ? new File([blob], file.name, { type: file.type }) : file),
+        file.type,
+        0.92,
+      );
+    };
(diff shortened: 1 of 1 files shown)


unitylibs/core/workflow/workflow-inline-action/action-binder.js
```diff
@@ -22,6 +22,60 @@ const UPLOAD_ERROR_TYPE = '.icon-error-request';
...
+    if ((marker & 0xFF00) !== 0xFF00) break;
+    offset += 2 + segmentLength;
+  }
+  return null;
+}
+
+async function correctOrientation(file) {
+  if (!/image\/(jpeg|jpg)/i.test(file.type)) return file;
+  const { default: isDesktop } = await import(`${getUnityLibs()}/utils/device-detection.js`);
+  if (isDesktop()) return file;
```
vipu0303: I believe we should remove this check.
This can happen on desktop as well for the photos which have different orientation.
The orientation check below should suffice, if it is 1, we skip, otherwise we correct the orientation.

sanjayms01: Good point, I realize it can occur. I have removed the check.

vipu0303 (approved):

arugupta1992 (approved):

asonnalagi:
Looks fine in the branch URL.
Before:
image
After:
image

PR #835, opened 2026-06-25
[MWPW-199606]Fix intermittent issue of experiment data not fetched for A/B testing

Ruchika4:
## Problem

A race condition existed between `workflow-acrobat`'s page config initialization and Adobe Launch (`_satellite`) not being ready at page load. When the unity block initialized, `ensurePageConfig` was fired fire-and-forget, which fetched page config and immediately called `window._satellite.track(...)` to retrieve Target experiment data. Since `_satellite` loads asynchronously, this call would throw a `TypeError` if Launch hadn't initialized yet. The error was caught and swallowed, leaving `experimentData` permanently `null` for that session because `pageConfigFetched = true` was already set, preventing any retry when the user actually uploaded a file.

### `unitylibs/utils/experiment-provider.js`
- Added `waitForSatellite()`  polls `window._satellite` every 100ms (up to 5 seconds) before invoking `_satellite.track(...)`
- Resolves immediately if `_satellite` is already present, rejects with a clear error on timeout
- Handles the common case where Launch initializes within a few seconds of page load

### `unitylibs/core/workflow/workflow-acrobat/action-binder.js`
- Refactored `ensurePageConfig` to handle the edge case where `_satellite` takes longer than 5 seconds or fails for other reasons
- On `getExperimentData` failure: reset `pageConfigFetched = false` and `pageConfigPromise = null` so the existing `|| this.ensurePageConfig()` guard in `handleFileUpload` naturally triggers a fresh attempt when the user uploads
- Added `const isRetry = !!this.pageConfigLocation` to detect first attempt vs user-triggered retry error is only dispatched to Splunk on retry, suppressing noise from the recoverable first-attempt failure
- `fetchPageConfig` catches all errors internally and always returns `{}`, so the catch block in `ensurePageConfig` is only ever reached via `getExperimentData` failure. Error is only dispatched to Splunk on retry (`isRetry`) since the first attempt failure is expected `_satellite` may not be loaded yet at page load. Flags are reset on failure to allow `ensurePageConfig` to be retried naturally when the user uploads a file.

### `test/utils/experiment-provider.test.js` / `test/core/workflow/workflow-acrobat/action-binder.test.js`
- Added 9 unit tests covering `waitForSatellite` polling and timeout, `ensurePageConfig` guard, flag reset without Splunk dispatch on first failure, flag reset with Splunk dispatch on retry, and success paths via both `pageConfig.config.target` and `targetCfg.experimentationOn`

## Behavior After Fix

| Scenario | Outcome |
|---|---|
| `_satellite` ready before page config completes | Experiment data fetched at page load, no retry needed |
| `_satellite` not ready within 5s | Silent failure at page load, retry triggered automatically when user uploads |
| Retry also fails | Error dispatched to Splunk, resets again for next upload attempt |

## Test plan
- [ ] Verify experiment data is correctly fetched on pages where `_satellite` loads before user interaction (happy path)
- [ ] Verify no visible error toast appears to users when experiment data fetch fails at page load
- [ ] Verify file upload works correctly when `_satellite` is delayed past the 5s timeout
- [ ] Verify `warn_fetch_experiment` Splunk entry is logged only on retry failure, not on initial page load failure

Resolves: [MWPW-199606](https://jira.corp.adobe.com/browse/MWPW-199606)

**Test URLs:**
- Before: https://main--da-dc--adobecom.aem.live/acrobat/online/split-pdf?unitylibs=stage
- After: https://main--da-dc--adobecom.aem.live/acrobat/online/split-pdf?unitylibs=page-config

### DIFF
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -273,11 +273,21 @@ export default class ActionBinder {
     if (this.pageConfigFetched) return;
     this.pageConfigFetched = true;
     const verb = this.workflowCfg.enabledFeatures[0];
+    let pageConfig;
     try {
       const { fetchPageConfig } = await import('../../../scripts/utils.js');
-      const { default: getExperimentData } = await import('../../../utils/experiment-provider.js');
-      const pageConfig = await fetchPageConfig({ product: 'acrobat', verb });
+      pageConfig = await fetchPageConfig({ product: 'acrobat', verb });
       this.pageConfigLocation = pageConfig.location;
+    } catch (error) {
+      await this.dispatchErrorToast('warn_fetch_experiment', null, error.message, true, true, {
+        code: 'warn_fetch_experiment',
+        desc: error.message,
+      });
+      this.acrobatApiConfig = this.getAcrobatApiConfig();
+      return;
+    }
+    try {
+      const { default: getExperimentData } = await import('../../../utils/experiment-provider.js');
       if (pageConfig.config?.target?.enabled) {
         this.experimentData = await getExperimentData(pageConfig.config.target.decisionScopes);
         this.experimentViaPageConfig = true;
@@ -291,6 +301,8 @@ export default class ActionBinder {
         code: 'warn_fetch_experiment',
         desc: error.message,
       });
+      this.pageConfigFetched = false;
+      this.pageConfigPromise = null;
     }
     this.acrobatApiConfig = this.getAcrobatApiConfig();
   }
--- a/unitylibs/core/workflow/workflow-acrobat/action-binder.js
+++ b/unitylibs/core/workflow/workflow-acrobat/action-binder.js
@@ -194,7 +194,6 @@ export default class ActionBinder {
     this.initialize();
     this.experimentData = null;
     this.experimentViaPageConfig = false;
-    this.experimentDataPromise = null;
     this.pageConfigLocation = null;
     this.pageConfigFetched = false;
     this.pageConfigPromise = null;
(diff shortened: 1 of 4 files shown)


unitylibs/core/workflow/workflow-acrobat/action-binder.js
```diff
@@ -287,10 +289,22 @@ export default class ActionBinder {
...
+          code: 'warn_fetch_experiment',
+          desc: error.message,
+        });
+        this.acrobatApiConfig = this.getAcrobatApiConfig();
+        return;
+      }
+      if (isRetry) {
+        await this.dispatchErrorToast('warn_fetch_experiment', null, error.message, true, true, {
+          code: 'warn_fetch_experiment',
+          desc: error.message,
```
vipu0303, line 295: do we need to dispatch the error toast in both the cases?

vipu0303 (approved):

unitylibs/core/workflow/workflow-acrobat/action-binder.js
Ruchika4: Good catch, Vipul. I realized that fetchPageConfig catches all errors internally and always returns {} (https://github.com/adobecom/unity/blob/stage/unitylibs/scripts/utils.js#L351), so the catch block in ensurePageConfig is only ever reached via getExperimentData failure. The isRetry check is to make sure we log the error only when even after the retry target data could not be fetched since at first attempt at page load there is a chance that launch is not ready and also avoid duplicate error messages to be logged. I have removed the check of fetchpageConfig from catch and now logging error only for isRetry.

shasmish (approved):

PR #836, opened 2026-07-06
Doodlebug Fast follow tickets 

arugupta1992:
[MWPW-195943](https://jira.corp.adobe.com/browse/MWPW-195943)
[MWPW-194444](https://jira.corp.adobe.com/browse/MWPW-194444)
[MWPW-193717](https://jira.corp.adobe.com/browse/MWPW-193717)
[MWPW-189974](https://jira.corp.adobe.com/browse/MWPW-189974)

**Fix Explanantion for MWPW-189974**
progressBarHandler used recursive setTimeout chains with no cleanup. After cancel, timers kept running and could overwrite the bar on a second upload or after updateProgressBar(100).

**Fix (in transition-screen.js)**
- clearProgressBarHandler() — clears the pending timeout and bumps a generation counter so stale callbacks no-op.
- progressBarHandler — on initialize=true, clears before reset; each callback checks its captured generation before updating.
- splashVisibilityController(false) — clears on hide/cancel (upload cancel path already uses this).
- updateProgressBar(..., 100) — clears auto-progress when completion is set explicitly (covers continueInApp in upload, acrobat, etc.).

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.live/products/firefly/features/sticker-generator?unitylibs=stage
- After: https://main--da-cc--adobecom.aem.live/products/firefly/features/sticker-generator?unitylibs=fastfollows#

### DIFF
--- a/unitylibs/scripts/transition-screen.js
+++ b/unitylibs/scripts/transition-screen.js
@@ -87,6 +87,7 @@ export default class TransitionScreen {
       this.progressBarTimeoutId = null;
       if (generation !== this.progressBarGeneration) return;
       const v = initialize ? 0 : parseInt(progressBar.getAttribute('value'), 10);
+      if (v === 100) return;
       this.updateProgressBar(s, v + newI);
       this.progressBarHandler(s, newDelay, newI);
     }, newDelay);
--- a/unitylibs/scripts/transition-screen.js
+++ b/unitylibs/scripts/transition-screen.js
@@ -25,7 +25,7 @@ export default class TransitionScreen {
   }
 
   clearProgressBarHandler() {
-    if (this.progressBarTimeoutId != null) {
+    if (this.progressBarTimeoutId !== null) {
       clearTimeout(this.progressBarTimeoutId);
       this.progressBarTimeoutId = null;
     }
@@ -87,7 +87,6 @@ export default class TransitionScreen {
       this.progressBarTimeoutId = null;
       if (generation !== this.progressBarGeneration) return;
       const v = initialize ? 0 : parseInt(progressBar.getAttribute('value'), 10);
-      if (v === 100) return;
       this.updateProgressBar(s, v + newI);
       this.progressBarHandler(s, newDelay, newI);
     }, newDelay);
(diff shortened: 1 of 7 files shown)


Ruchika4:
Hi @arugupta1992 , since the PR has changes in common file transition-screen.js which is used for transition screen across all workflows, I am adding label platform so that a quick sanity can be done across all workflows. cc @sigadamvenkata @asonnalagi

unitylibs/scripts/transition-screen.js
```diff
@@ -20,6 +20,16 @@ export default class TransitionScreen {
     this.isDesktop = isDesktop;
     this.headingElements = [];
     this.progressText = '';
+    this.progressBarTimeoutId = null;
+    this.progressBarGeneration = 0;
+  }
+
+  clearProgressBarHandler() {
+    if (this.progressBarTimeoutId != null) {
```
Ruchika4: Nit: consider using !== null here. != null works correctly in this case but strict equality is more explicit and avoids any potential eqeqeq lint rule failures.
```suggestion
    if (this.progressBarTimeoutId !== null) {
```

unitylibs/scripts/transition-screen.js
```diff
@@ -63,13 +74,18 @@ export default class TransitionScreen {
...
       if (currentValue === 100 || currentValue >= this.LOADER_LIMIT) return;
     }
 
-    setTimeout(() => {
+    const generation = this.progressBarGeneration;
+    this.progressBarTimeoutId = setTimeout(() => {
+      this.progressBarTimeoutId = null;
+      if (generation !== this.progressBarGeneration) return;
       const v = initialize ? 0 : parseInt(progressBar.getAttribute('value'), 10);
       if (v === 100) return;
```
Ruchika4, line 90: Nit: the `if (v === 100) return` inside the setTimeout callback appears to be dead code post this update. if v were 100, updateProgressBar would have already called clearProgressBarHandler(), causing the generation check above it to exit first. Please check once and see if above understanding is correct and if we need to remove it.

vipu0303 (approved):

unitylibs/scripts/transition-screen.js
arugupta1992: Thanks for catching this. 
This is being used for Acrobat workflow. In acrobat a new instance is created for transition screen in continueInApp to make it 100%. So on removing this, the progress barr moves to 100% (new instance) and then then the old instance progress pops and 95% is shown on UI

asonnalagi:
Looks fine in branch. Test results are in the JIRA's.

PR #846, opened 2026-07-14
Progress bar is not filled till the end even on 100% for acrobat

arugupta1992:
RC: 

1. Multiple TransitionScreen instances shared the same splash DOM — loadTransitionScreen(), upload-handler.initSplashScreen(), and continueInApp() each created their own instance. Auto-progress timers are per-instance, so orphaned timers from earlier instances could keep updating the bar after the final 100% state was set.
2. Label vs fill mismatch — the percentage text updated instantly, but the fill used a 1s CSS width transition, while redirect happened after only 500ms, so the bar often looked incomplete at redirect time.

**Test URLs:**
- Before: https://main--da-dc--adobecom.aem.live/acrobat/online/sign-pdf?unitylibs=stage
- After: https://main--da-dc--adobecom.aem.live/acrobat/online/sign-pdf?unitylibs=fastfollows3

### DIFF
--- a/test/core/workflow/workflow-acrobat/action-binder.test.js
+++ b/test/core/workflow/workflow-acrobat/action-binder.test.js
@@ -1597,6 +1597,7 @@ describe('ActionBinder', () => {
         const existingTransitionScreen = {
           splashScreenEl: splashLayer,
           LOADER_LIMIT: 95,
+          clearProgressBarHandler: sinon.stub(),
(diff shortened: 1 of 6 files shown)


asonnalagi:
Verified in the branch. Looks fine.
https://main--da-dc--adobecom.aem.live/acrobat/online/sign-pdf?unitylibs=fastfollows3
image

